PR #22, opened 2024-11-27
feat: P4ADEV-1545 fdr ingestion actity implementation retrieval via query file validation

macacia:
#### Description
Implementation of the following steps:
- validate the file (it exists, md5)
- decrypt (in temporary directory)
- zip security checks? (is it a zip?)
- possible unzip (in temporary directory)

- Pre-Deploy Test
    - [x] Unit
    - [ ] Integration (Narrow)
- Post-Deploy Test
    - [ ] Isolated Microservice
    - [ ] Broader Integration
    - [ ] Acceptance
    - [ ] Performance & Load

#### Types of changes

#### Checklist:


src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/IngestionFileHandlerService.java
```diff
@@ -0,0 +1,71 @@
...
+import java.nio.file.Paths;
+
+/**
+ * Service class responsible for handling encrypted ingestion files,
+ * including decryption, validation, and extraction of ZIP files.
+ */
+@Slf4j
+@Service
+public class IngestionFileHandlerService {
+	private static final String TEMPORARY_PATH = "TEMP";
```
antonioT90, line 21: This way you are referencing a local directory, not the tmp directory...
The local directory probably it will be the share.

maybe it's better to use the /tmp/ directory of the running pod in order to avoid to store in the share personal data
(you have to create a folder based on the input file, maybe under this folder you could create the same folder hierarchy of the current file)

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/IngestionFileHandlerService.java
```diff
@@ -0,0 +1,71 @@
...
+	 * Handles the setup process for an ingestion file by performing the following steps:
+	 * <ul>
+	 *     <li>Decrypts the file using AES encryption.</li>
+	 *     <li>Validates if the file is a valid ZIP archive.</li>
+	 *     <li>Extracts the contents of the ZIP file to a temporary directory.</li>
+	 * </ul>
+	 *
+	 * @param relativePath the relative path to the directory containing the file.
+	 * @param filename the name of the file to process.
+	 * @return the path to the extracted XML file.
```
antonioT90, line 42: This will be a generic service which will decipher and uncompress the input file, whatever is it's extension
```suggestion
	 * @return the path to the extracted file.
```

src/main/java/it/gov/pagopa/payhub/activities/util/FileUtils.java
```diff
@@ -0,0 +1,194 @@
...
+			}
+		}
+	}
+
+	/**
+	 * Protects against ZIP Slip vulnerability by validating the path of an extracted entry.
+	 *
+	 * @param zipEntry the ZIP entry to validate.
+	 * @param targetDir the target directory for extraction.
+	 * @return a safe, normalized path within the target directory.
```
antonioT90, line 182:
```suggestion
	 * @return a safe, normalized path towards zipEntry within the target directory.
```

src/main/java/it/gov/pagopa/payhub/activities/util/FileUtils.java
```diff
@@ -0,0 +1,194 @@
...
+	 *                                        <ul>
+	 *                                          <li>The ZIP file contains an entry with an invalid name
+	 *                                          that could lead to a ZIP Slip attack.</li>
+	 *                                          <li>The ZIP file exceeds the maximum allowed uncompressed size.</li>
+	 *                                          <li>The ZIP file contains an excessive number of entries.</li>
+	 *                                          <li>An entry in the ZIP file has a suspiciously high compression ratio.</li>
+	 *                                          <li>An I/O error occurs during extraction.</li>
+	 *                                        </ul>
+	 */
+	public static void unzip(Path source, Path target) {
```
antonioT90, line 91: this method could return the list of unzipped Path

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/IngestionFileHandlerService.java
```diff
@@ -0,0 +1,71 @@
...
+		String filenameNoCipher = filename.replace(AESUtils.CIPHER_EXTENSION, "");
+		Path temporaryZipFilePath = temporaryPath.resolve(filenameNoCipher);
+
+		log.debug("Decrypting file: {}", encryptedFilePath);
+		AESUtils.decrypt(dataCipherPsw, encryptedFilePath.toFile(), temporaryZipFilePath.toFile());
+
+		log.debug("Validating ZIP file: {}", temporaryZipFilePath);
+		FileUtils.isArchive(temporaryZipFilePath);
+
+		String unzippedFilename = filenameNoCipher.replace(".zip", ".xml");
```
antonioT90, line 62: FileUtils.unzip will accept a directory, not the final unzipped filename
thus you have to call the unzip with just the temporaryPath, using its returned value to retrieve the Path to the unzipped file (you have to modify unzip method to return the list of unzip files

we could let this setupProcess to return a List or (letting the invoking method to validate the presence of just 1 file) or set here the check of the unzipped file number

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/IngestionFileHandlerService.java
```diff
@@ -0,0 +1,71 @@
...
+	 * @param filename the name of the file to process.
+	 * @return the path to the extracted XML file.
+	 * @throws IOException if any file operation fails during the setup process.
+	 */
+	public Path setUpProcess(String relativePath, String filename) throws IOException {
+		Path relativePathDir = Paths.get(relativePath);
+		Path encryptedFilePath = relativePathDir.resolve(filename);
+		FileUtils.validateFile(encryptedFilePath);
+
+		Path temporaryPath = relativePathDir.resolve(TEMPORARY_PATH);
```
antonioT90, line 50: if TEMPORARY_PATH will be `Paths.of("/tmp")`, this could be
TEMPORARY_PATH.resolve(relativePathDir.subpath(0, relativePathDir.getNameCount()))

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/IngestionFileHandlerService.java
```diff
@@ -0,0 +1,71 @@
...
+	 *     <li>Validates if the file is a valid ZIP archive.</li>
+	 *     <li>Extracts the contents of the ZIP file to a temporary directory.</li>
+	 * </ul>
+	 *
+	 * @param relativePath the relative path to the directory containing the file.
+	 * @param filename the name of the file to process.
+	 * @return the path to the extracted XML file.
+	 * @throws IOException if any file operation fails during the setup process.
+	 */
+	public Path setUpProcess(String relativePath, String filename) throws IOException {
```
antonioT90, line 45: This class could be just IngestionFileRetrieverService instead of a generic Handler
```suggestion
	public Path retrieveFile(String relativePath, String filename) throws IOException {
```

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/IngestionFileHandlerService.java
macacia: According to the legacy system, the temporary folder was always at the end and not at the root

antonioT90: The change proposed is not a business logic change, but a technical requirement.
The current directory is inside a share, let's avoid to overuse it in order to:
* Reduce the network overhead
* Avoid to expose plain data
* Avoid to dirty the folder with temporary files of interest of just the running pod

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/IngestionFlowRetrieverService.java
antonioT90, line 1: This Service is ambiguous considering the existence of the sibling class on this same packge...
let's remove this class and instead let's call directly the DAO (the only value is the not found handling...)

src/test/java/it/gov/pagopa/payhub/activities/util/FilesUtils.java
```diff
@@ -0,0 +1,15 @@
+package it.gov.pagopa.payhub.activities.util;
+
+import java.io.IOException;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipOutputStream;
+
+public class FilesUtils {
```
antonioT90, line 7: is this method really generic? it seems to me to be a method which will not used outside the context of a zip file... let's move this method on the class itself, removing this class

src/test/java/it/gov/pagopa/payhub/activities/service/ingestionflow/IngestionFileRetrieverServiceTest.java
```diff
@@ -0,0 +1,116 @@
...
+	@Mock
+	private ZipFileService zipFileService;
+
+	@InjectMocks
+	private IngestionFileRetrieverService ingestionFileRetrieverService;
+
+	private Path zipFile;
+
+	@TempDir
+	Path tempDir;
```
antonioT90, line 43:
```suggestion
	private Path tempDir;
```

antonioT90 (changes requested):

antonioT90 (dismissed):

antonioT90 (approved):

antonioT90 (approved):

PR #27, opened 2024-11-28
feat: P4ADEV-1544 reportingflow ingestion actity implementation archiving and cleaning operation

macacia:

- Pre-Deploy Test
    - [x] Unit
    - [ ] Integration (Narrow)
- Post-Deploy Test
    - [ ] Isolated Microservice
    - [ ] Broader Integration
    - [ ] Acceptance
    - [ ] Performance & Load

#### Types of changes

#### Checklist:


src/main/java/it/gov/pagopa/payhub/activities/activity/paymentsreporting/PaymentsReportingIngestionFlowFileActivityImpl.java
```diff
@@ -34,33 +35,37 @@ public class PaymentsReportingIngestionFlowFileActivityImpl implements PaymentsR
...
 
-			File ingestionFlowFile = retrieveFile(ingestionFlowFileDTO);
+			List<Path> filePaths = retrieveFile(ingestionFlowFileDTO);
 
-			Pair<String, List<PaymentsReportingDTO>> pair = parseData(ingestionFlowFile, ingestionFlowFileDTO);
+			Pair<String, List<PaymentsReportingDTO>> pair = parseData(filePaths.get(0).toFile(), ingestionFlowFileDTO);
 			paymentsReportingDao.saveAll(pair.getRight());
-
+			ingestionFlowFileAchiverService
+				.compressArchiveFileAndCleanUp(filePaths, Path.of(ingestionFlowFileDTO.getFilePathName()), ingestionFlowFileDTO.getFileName());
```
antonioT90, line 68: This is the you ciphering, then compressing the file that originally you have decompressed and ciphered, why don't you just move the original file instead of re-creating it?

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/IngestionFlowFileAchiverService.java
```diff
@@ -0,0 +1,69 @@
...
+	private final String targetDirectory;
+
+	/**
+	 * The password used for encrypting files.
+	 */
+	private final String dataCipherPsw;
+
+	private final ZipFileService zipFileService;
+
+	public IngestionFlowFileAchiverService(@Value("${stored-dir:/processed/}") String targetDirectory,
```
antonioT90, line 36: the target directory should be on the shared folder as the original file.
should the target directory be a subdirectory of the original position?

src/main/java/it/gov/pagopa/payhub/activities/activity/paymentsreporting/PaymentsReportingIngestionFlowFileActivityImpl.java
```diff
@@ -34,33 +35,37 @@ public class PaymentsReportingIngestionFlowFileActivityImpl implements PaymentsR
...
 	@Override
 	public PaymentsReportingIngestionFlowFileActivityResult processFile(Long ingestionFlowFileId) {
 		try {
 			IngestionFlowFileDTO ingestionFlowFileDTO = findIngestionFlowFileRecord(ingestionFlowFileId);
 
-			File ingestionFlowFile = retrieveFile(ingestionFlowFileDTO);
+			List<Path> filePaths = retrieveFile(ingestionFlowFileDTO);
 
-			Pair<String, List<PaymentsReportingDTO>> pair = parseData(ingestionFlowFile, ingestionFlowFileDTO);
+			Pair<String, List<PaymentsReportingDTO>> pair = parseData(filePaths.get(0).toFile(), ingestionFlowFileDTO);
```
antonioT90, line 65: the original process is creating a file where to store the errors at row level?

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/IngestionFlowFileAchiverService.java
```diff
@@ -0,0 +1,69 @@
...
+
+import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
+
+/**
+ * Service class responsible for handling archiving ingestion files,
+ * including zipping and ecryption files.
+ */
+@Lazy
+@Service
+public class IngestionFlowFileAchiverService {
```
antonioT90, line 23:
```suggestion
public class IngestionFlowFileArchiverService {
```

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/IngestionFlowFileAchiverService.java
macacia: We will use the original file path, the target dir will be a subfolder

src/main/java/it/gov/pagopa/payhub/activities/activity/paymentsreporting/PaymentsReportingIngestionFlowFileActivityImpl.java
macacia: in the legacy solution it was not foreseen, due to the fact that it is all ok or nothing and the only info data is returned to the FE which is displayed in the UI

src/main/java/it/gov/pagopa/payhub/activities/activity/paymentsreporting/PaymentsReportingIngestionFlowFileActivityImpl.java
```diff
@@ -91,13 +98,12 @@ private IngestionFlowFileDTO findIngestionFlowFileRecord(Long ingestionFlowFileI
 	 * extracting it from the specified file path.
 	 *
 	 * @param ingestionFlowFileDTO the ingestion flow file DTO containing file details
-	 * @return the extracted {@link File} from the ingestion flow
+	 * @return the extracted {@link List} from the ingestion flow
 	 * @throws IOException if there is an error during file retrieval or extraction
 	 */
-	private File retrieveFile(IngestionFlowFileDTO ingestionFlowFileDTO) throws IOException {
-		List<Path> ingestionFlowFiles = ingestionFlowFileRetrieverService
+	private List<Path> retrieveFile(IngestionFlowFileDTO ingestionFlowFileDTO) throws IOException {
```
antonioT90, line 104: Why are you changing from a single Path to a List?
If the validation see more than 1 file it would throw an exception (is the validation removing the file in that case?)?

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/IngestionFlowFileArchiverService.java
```diff
@@ -0,0 +1,69 @@
...
+	 * Creates the target directory if it does not exist.
+	 *
+	 * @param files2Archive the list of files to move to the target directory.
+	 * @param targetPath the directory where the files will be moved.
+	 * @throws IOException if an error occurs during directory creation, file movement, or cleanup.
+	 */
+	public void archive(List<Path> files2Archive, Path targetPath) throws IOException {
+		Files.createDirectories(targetPath);
+		for (Path file : files2Archive) {
+			Files.copy(file, targetPath.resolve(file.toFile().getName()), REPLACE_EXISTING);
```
antonioT90, line 64: Let's not create a File object just to retrieve file name
```suggestion
			Files.copy(file, targetPath.resolve(file.getFileName()), REPLACE_EXISTING);
```

src/main/java/it/gov/pagopa/payhub/activities/activity/paymentsreporting/PaymentsReportingIngestionFlowFileActivityImpl.java
```diff
@@ -28,38 +30,46 @@
...
+			File retrievedFile = retrieveFile(ingestionFlowFileDTO);
+
+			Pair<String, List<PaymentsReportingDTO>> pair = parseData(retrievedFile, ingestionFlowFileDTO);
 
-			Pair<String, List<PaymentsReportingDTO>> pair = parseData(ingestionFlowFile, ingestionFlowFileDTO);
 			paymentsReportingDao.saveAll(pair.getRight());
+			archive(ingestionFlowFileDTO);
 
 			return new PaymentsReportingIngestionFlowFileActivityResult(List.of(pair.getLeft()), true);
 		} catch (Exception e) {
```
antonioT90, line 75: if an exception occurred, you should delete the retrievedFile if notNull

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/activity/paymentsreporting/PaymentsReportingIngestionFlowFileActivityImpl.java
```diff
@@ -118,4 +128,17 @@ private Pair<String, List<PaymentsReportingDTO>> parseData(File ingestionFlowFil
...
+
+	/**
+	 * Archives the file specified in the given {@link IngestionFlowFileDTO}. The file is moved to
+	 * the archive directory located within the same file path.
+	 *
+	 * @param ingestionFlowFileDTO the DTO containing details of the file to be archived.
+	 * @throws IOException if an error occurs during file movement or directory creation.
+	 */
+	private void archive(IngestionFlowFileDTO ingestionFlowFileDTO) throws IOException {
+		Path originalFilePath = Paths.get(ingestionFlowFileDTO.getFilePathName(), ingestionFlowFileDTO.getFileName());
```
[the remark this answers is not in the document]
macacia: dto field renamed

src/main/java/it/gov/pagopa/payhub/activities/activity/paymentsreporting/PaymentsReportingIngestionFlowFileActivityImpl.java
```diff
@@ -19,52 +20,67 @@
...
 		this.paymentsReportingIngestionFlowFileValidatorService = paymentsReportingIngestionFlowFileValidatorService;
 		this.paymentsReportingMapperService = paymentsReportingMapperService;
 		this.paymentsReportingDao = paymentsReportingDao;
+		this.ingestionFlowFileArchiverService = ingestionFlowFileArchiverService;
 	}
 
 	@Override
-	public PaymentsReportingIngestionFlowFileActivityResult processFile(Long ingestionFlowFileId) {
+	public PaymentsReportingIngestionFlowFileActivityResult processFile(Long ingestionFlowFileId) throws IOException {
+		Optional<File> retrievedFile = Optional.empty();
```
antonioT90, line 66: let's not use Optional, we should handle the null just on the catch clause, we have not value into introducing it:
the catch is the only place where it could be null, if you use Optional, then you should handle the ifPresence where now you are simply executing the get

src/main/java/it/gov/pagopa/payhub/activities/activity/paymentsreporting/PaymentsReportingIngestionFlowFileActivity.java
```diff
@@ -13,5 +15,5 @@ public interface PaymentsReportingIngestionFlowFileActivity {
      * @param ingestionFlowFileId the unique identifier related to the file to process.
      * @return {@link PaymentsReportingIngestionFlowFileActivityResult} containing the list of IUFs and status.
      */
-    PaymentsReportingIngestionFlowFileActivityResult processFile(Long ingestionFlowFileId);
+    PaymentsReportingIngestionFlowFileActivityResult processFile(Long ingestionFlowFileId) throws IOException;
```
antonioT90, line 18: let's eliminate this throws clause

PR #40, opened 2024-12-04
feat: P4ADEV-1373 generate iuv

serdimic:
#### Description

Implement generate IUV activity.

**_Remark_**: this activity should not be used "standalone", but inside the activity creating a debt position; this code will be subject to refactoring. Therefore, as agreed, there is currently no transaction management; it will be taken care of during refactoring.

#### List of Changes

Add generate IUV activity and related service/dao/test classes.

#### Motivation and Context

Implement generate IUV activity.

- Pre-Deploy Test
    - [X] Unit
    - [ ] Integration (Narrow)
- Post-Deploy Test
    - [ ] Isolated Microservice
    - [ ] Broader Integration
    - [ ] Acceptance
    - [ ] Performance & Load

#### Types of changes

#### Checklist:


src/main/java/it/gov/pagopa/payhub/activities/dao/IuvSequenceNumberDao.java
```diff
@@ -0,0 +1,12 @@
+package it.gov.pagopa.payhub.activities.dao;
+
+public interface IuvSequenceNumberDao {
```
antonioT90, line 3: Can you help me to locate this entity on the mapping document?
https://pagopa.atlassian.net/wiki/spaces/SPAC/pages/1406238722/Mapping+entit

src/main/java/it/gov/pagopa/payhub/activities/service/IuvService.java
```diff
@@ -0,0 +1,109 @@
...
+@Service
+@Slf4j
+public class IuvService {
+
+  private static final String PROGRESSIVO_PADDING_ZEROES_SMALL = "%011d";
+  public static final String AUX_DIGIT = "3";
+
+  private final String informationSystemId;
+
+  private final IuvSequenceNumberDao progressiviVersamentoDao;
```
antonioT90, line 25: lets align variable and class name (use english)

src/main/java/it/gov/pagopa/payhub/activities/service/IuvService.java
```diff
@@ -0,0 +1,109 @@
...
+
+  public IuvService(@Value("${iuv.informationSystemId:00}") String informationSystemId,
```
antonioT90, line 27: What is the `informationSystemId`?

src/main/java/it/gov/pagopa/payhub/activities/service/IuvService.java
```diff
@@ -0,0 +1,109 @@
...
+  public String generateIuv(OrganizationDTO org){
+    StringBuilder iuvBuilder = new StringBuilder();
+    iuvBuilder.append(org.getApplicationCode());
+    iuvBuilder.append(informationSystemId);
+    long paymentIndex = progressiviVersamentoDao.getNextIuvSequenceNumber(org.getIpaCode());
+    if(paymentIndex<1){
+      log.error("invalid payment index returned for org[{}/{}]: {}", org.getIpaCode(), org.getOrgFiscalCode(), paymentIndex);
+      throw new ValueNotValidException("invalid payment index");
+    }
+    iuvBuilder.append(String.format(PROGRESSIVO_PADDING_ZEROES_SMALL, paymentIndex));
```
antonioT90, line 48: why don't you use the same function to perform leftPad as done on row 54?

src/main/java/it/gov/pagopa/payhub/activities/service/IuvService.java
```diff
@@ -0,0 +1,109 @@
...
+
+    //check digit
+    String digitString = AUX_DIGIT + iuvBuilder;
+    long digit = Long.parseLong(digitString);
+    long reminder = digit % 93;
+    String checkDigit = StringUtils.leftPad(String.valueOf(reminder), 2, '0');
```
antonioT90, line 54: let's move on a separate method the checkDigit building logic, this way the main method could contains just the append operations, keeping it more readable

src/main/java/it/gov/pagopa/payhub/activities/service/IuvService.java
antonioT90, line 48: let's move paymentIndex (also padded) building logic on a separate method, in order to let more readable the main method

src/main/java/it/gov/pagopa/payhub/activities/service/IuvService.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+  }
+
+  /**
+   * Utility method to formally validate a IUV.
+   * @param iuv the IUV to validate
+   * @return true if valid, otherwise false
+   */
+  public boolean isValidIuv(String iuv){
+    return isValidNav(StringUtils.join(AUX_DIGIT+iuv));
```
antonioT90, line 91: the join method is not really used, just use the simple concat operation.
Moreover, you are performing this concatenation also at row 67, why don't you externalize it on a private method `innerIuv2Nav` instead of replicating it?

src/main/java/it/gov/pagopa/payhub/activities/service/IuvService.java
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * Utility method to formally validate a NAV.
+   * @param nav the NAV to validate
+   * @return true if valid, otherwise false
+   */
+  public boolean isValidNav(String nav){
+    if(StringUtils.length(nav)==18 && StringUtils.startsWith(nav, AUX_DIGIT)){
+      try{
+        return Long.parseLong(nav.substring(0,16)) % 93 == Long.parseLong(nav.substring(16));
```
antonioT90, line 102: I see you used this 93 also on checkDigit, why don't you store it on a constant variable in order to centralize its content?

src/main/java/it/gov/pagopa/payhub/activities/service/OrganizationService.java
```diff
@@ -0,0 +1,31 @@
...
+    this.organizationDao = organizationDao;
+  }
+
+  /*
+   * Retrieve organization entity by fiscal code.
+   * @param orgFiscalCode the fiscal code of the organization to retrieve
+   * @return the retrieved organization
+   */
+  public Optional<OrganizationDTO> getOrganizationByFiscalCode(String orgFiscalCode){
+    return Optional.ofNullable(organizationDao.getOrganizationByFiscalCode(orgFiscalCode));
```
antonioT90, line 29: we could change the DAO to return an Optional instead of just organizationDTO if we want to handle it always as an optional instead of moving this logic on the Service (SpringData is able to handle Optional, so this behavior would be handled OutOfTheBox by the framework on the new microservices)

src/main/java/it/gov/pagopa/payhub/activities/exception/ValueNotValidException.java
```diff
@@ -0,0 +1,9 @@
+package it.gov.pagopa.payhub.activities.exception;
+
+
+public class ValueNotValidException extends ActivitiesException {
```
antonioT90, line 4: I see we have already implemented many exceptions to represent not valid data, but there we used the form Invalid<kind of data>Exception
```suggestion
public class InvalidValueException extends ActivitiesException {
```

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/service/IuvService.java
serdimic: The "information system id" is the two-digit code that identifies an information system that could generate IUV for the organization / segregationCode. Since this code is used when composing the IUV (in position 3 and 4 of the generated IUV), it ensures that IUV generated by different information systems will not collide.
From the point of view of IUV, MyPay / PU is just one of the information systems generating IUV, and by default has 00 as its informationSystemId.
This means that:
- every IUV generated by MyPay / PU will have 00 as the 3° and 4° digit;
- every IUV generated externally (i.e. NOT by MyPay) will have 3° and 4° digit != '00';
- if there are multiple external information systems generating IUV for this ente, any of them will use its informationSystemId.

src/main/java/it/gov/pagopa/payhub/activities/dao/IuvSequenceNumberDao.java
serdimic: done

src/main/java/it/gov/pagopa/payhub/activities/service/IuvService.java
serdimic: fixed

src/main/java/it/gov/pagopa/payhub/activities/service/IuvService.java
serdimic: I just copy-pasted the code from MyPay; fixed now.

src/main/java/it/gov/pagopa/payhub/activities/service/IuvService.java
serdimic: done

src/main/java/it/gov/pagopa/payhub/activities/service/IuvService.java
serdimic: done

src/main/java/it/gov/pagopa/payhub/activities/service/IuvService.java
serdimic: done

src/main/java/it/gov/pagopa/payhub/activities/service/OrganizationService.java
serdimic: done

src/main/java/it/gov/pagopa/payhub/activities/service/IuvService.java
serdimic: There was a typo that I fixed, the correct join is:
`isValidNav(StringUtils.join(AUX_DIGIT,iuv))`
I would prefer keeping the `join` function since it better handle the null (it will prevent having `"a"+null` resulting as an `"anull"` string).
Also while apparently this is the same operation used inside `iuv2Nav` method:
```
    if(isValidIuv(iuv))
      return AUX_DIGIT + iuv;
```
the context is different in a subtle way: here the concat is used on an already validated iuv so there is no risk of having null, therefore there is no need to use the join.

src/main/java/it/gov/pagopa/payhub/activities/dao/IuvSequenceNumberDao.java
antonioT90: Where have you reported this mapping? I cannot see it

serdimic: because i didn't publish the modifications... done now

src/main/java/it/gov/pagopa/payhub/activities/service/IuvService.java
antonioT90: is it right to configure it as a global variable or it should change based on the broker?

antonioT90 (changes requested):
even if reported for test files, could we resolve Sonar issues?

src/main/java/it/gov/pagopa/payhub/activities/service/IuvService.java
serdimic: Considering that currently on MyPay the value of "information System id" is hardcoded in the code, I would keep the handling of this value as simple as possible for the moment. If in future there is the need (but I doubt it) to handle different values per broker, we may always introduce this customization.

antonioT90 (approved):

PR #47, opened 2024-12-10
feat: P4ADEV-1656 Treasury Activity ingestion file

domenicogi:
#### Description

First part of treasury activity with integration of ingetion file flow

- Pre-Deploy Test
    - [X] Unit
    - [ ] Integration (Narrow)
- Post-Deploy Test
    - [ ] Isolated Microservice
    - [ ] Broader Integration
    - [ ] Acceptance
    - [ ] Performance & Load

#### Types of changes

#### Checklist:


src/main/java/it/gov/pagopa/payhub/activities/dto/treasury/TreasuryIngestionResultDTO.java
antonioT90, line 1: Why have you renamed this DTO? It was created with `Iuf` in order to distinguish between TreasuryIuv ingestion and TreasuryIuf ingestion.

OPI injestion could receive also IUV? As I know, just "Giornale di cassa XLS" could contains IUV

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
antonioT90, line 1: can you refactor in order to have a clean processFile method (moving on private methods the steps) as Marco done in PaymentsReporting?

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
```diff
@@ -0,0 +1,69 @@
...
+  public TreasuryIngestionResultDTO processFile(Long ingestionFlowFileId) {
+    List<IufIuvDTO> iufIuvList = new ArrayList<>();
+    try {
+      IngestionFlowFileDTO ingestionFlowFileDTO = ingestionFlowFileDao.findById(ingestionFlowFileId)
+              .orElseThrow(() -> new IngestionFlowFileNotFoundException("Cannot found ingestionFlow having id: " + ingestionFlowFileId));
+      if (!ingestionFlowFileDTO.getFlowFileType().equals(ingestionflowFileType)) {
+        throw new IllegalArgumentException("invalid ingestionFlow file type " + ingestionFlowFileDTO.getFlowFileType());
+      }
+
+      List<Path> ingestionFlowFiles = ingestionFlowFileRetrieverService
```
antonioT90, line 52: Do we expected more than one file?

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
domenicogi: Inside the zip file can be more than one xml file

src/main/java/it/gov/pagopa/payhub/activities/dto/treasury/TreasuryIngestionResultDTO.java
domenicogi: For Opi it's required only IUF but i made a general implementation for Treasury. If you prefer i can rollback to a simple list of string contains IUF

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
```diff
@@ -0,0 +1,80 @@
...
+@Slf4j
+@Lazy
+@Component
+public class TreasuryOpiIngestionActivityImpl implements TreasuryOpiIngestionActivity {
+  private final String ingestionflowFileType;
+  private final IngestionFlowFileDao ingestionFlowFileDao;
+  private final IngestionFlowFileRetrieverService ingestionFlowFileRetrieverService;
+
+
+  public TreasuryOpiIngestionActivityImpl(@Value("${ingestion-flow-file-type:O}") String ingestionflowFileType,
```
antonioT90, line 34: let's start using an Enum for this field
it will not be a variable configurable through properties

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
```diff
@@ -0,0 +1,80 @@
...
+    try {
+      IngestionFlowFileDTO ingestionFlowFileDTO = findIngestionFlowFileRecord(ingestionFlowFileId);
+
+      List<Path> ingestionFlowFiles = retrieveFiles(ingestionFlowFileDTO);
+
+      ingestionFlowFiles.forEach(path -> {
+        File ingestionFlowFile = path.toFile();
+        log.debug("file from zip archive with name {} loaded successfully ", ingestionFlowFile.getName());
+
+      });
```
antonioT90, line 55: don't create a File object just to read fileName, you could use Path#getFileName() instead.
Moreover, you are making a cycle just to print a debug log... if the actual level is not debug, you are looping for nothing (you could use log.isDebugEnabled to perform the action if needed).

Finally, for debug purporse, we could think to print the entire path, so at the very end it could be simplified as follow:
```suggestion
     log.debug("Successfully retrieved the following files related to the ingestionFlowFileId {}: {}", ingestionFlowFileId, ingestionFlowFiles);
```

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/dto/IngestionFlowFileDTO.java
```diff
@@ -32,4 +34,10 @@ public class IngestionFlowFileDTO implements Serializable {
     private Long pdfGenerated;
     private String codRequestToken;
     private String codError;
+    private String pspIdentifier;
+    private Date flowDateTime;
```
antonioT90, line 38: let's use new date/time API

are these columns on the table that Matteucci is creating?
```suggestion
    private LocalDateTime flowDateTime;
```

src/main/java/it/gov/pagopa/payhub/activities/dto/IngestionFlowFileDTO.java
```diff
@@ -32,4 +34,10 @@ public class IngestionFlowFileDTO implements Serializable {
...
+    private BigInteger state;
```
antonioT90, line 39: this is the status? if yes, I expect it should be a String or better an enum (actually could be String, we could introduce the enum after)

antonioT90 (changes requested):

PR #49, opened 2024-12-10
feat: P4ADEV-1657 treasury opi ingestion opi validation

domenicogi:
#### Description

Added opi validation in treasury activity

- Pre-Deploy Test
    - [X] Unit
    - [ ] Integration (Narrow)
- Post-Deploy Test
    - [ ] Isolated Microservice
    - [ ] Broader Integration
    - [ ] Acceptance
    - [ ] Performance & Load

#### Types of changes

#### Checklist:


src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
```diff
@@ -26,50 +28,87 @@
...
-                                          IngestionFlowFileRetrieverService ingestionFlowFileRetrieverService) {
-    this.ingestionflowFileType = IngestionFlowFileType.OPI;
-    this.ingestionFlowFileDao = ingestionFlowFileDao;
-    this.ingestionFlowFileRetrieverService = ingestionFlowFileRetrieverService;
-  }
+    @Override
+    public TreasuryIufResult processFile(Long ingestionFlowFileId) {
+        List<String> iufIuvList = new ArrayList<>();
+        List<Path> ingestionFlowFiles = null;
+        AtomicBoolean success = new AtomicBoolean(true);
```
antonioT90, line 52: AtomicBoolean is a thread safe blocking operation, don't use it if not really necessary.
You are processing rows one by one, if any goes into error the entire process should be discarded, so a variable is not needed

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
```diff
@@ -26,50 +28,87 @@
...
-  public TreasuryOpiIngestionActivityImpl(IngestionFlowFileDao ingestionFlowFileDao,
-                                          IngestionFlowFileRetrieverService ingestionFlowFileRetrieverService) {
-    this.ingestionflowFileType = IngestionFlowFileType.OPI;
-    this.ingestionFlowFileDao = ingestionFlowFileDao;
-    this.ingestionFlowFileRetrieverService = ingestionFlowFileRetrieverService;
-  }
+    @Override
+    public TreasuryIufResult processFile(Long ingestionFlowFileId) {
+        List<String> iufIuvList = new ArrayList<>();
+        List<Path> ingestionFlowFiles = null;
```
antonioT90, line 51: clean operations will be implemented in a separate task?

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
```diff
@@ -26,50 +28,87 @@
...
-    List<String> iufIuvList = new ArrayList<>();
-    try {
-      IngestionFlowFileDTO ingestionFlowFileDTO = findIngestionFlowFileRecord(ingestionFlowFileId);
+        } catch (Exception e) {
+            log.error("Error during TreasuryOpiIngestionActivity ingestionFlowFileId {}", ingestionFlowFileId, e);
+            return new TreasuryIufResult(Collections.emptyList(), false);
+        }
 
-      List<Path> ingestionFlowFiles = retrieveFiles(ingestionFlowFileDTO);
+        if (ingestionFlowFiles != null && !ingestionFlowFiles.isEmpty()) {
```
antonioT90, line 63: ingestionFlowFileRetrieverService could return null?

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
```diff
@@ -26,50 +28,87 @@
...
 
-  @Override
-  public TreasuryIufResult processFile(Long ingestionFlowFileId) {
-    List<String> iufIuvList = new ArrayList<>();
-    try {
-      IngestionFlowFileDTO ingestionFlowFileDTO = findIngestionFlowFileRecord(ingestionFlowFileId);
+        } catch (Exception e) {
+            log.error("Error during TreasuryOpiIngestionActivity ingestionFlowFileId {}", ingestionFlowFileId, e);
+            return new TreasuryIufResult(Collections.emptyList(), false);
+        }
```
antonioT90, line 61: why this try/catch is just for the above lines and not also the parse method? (could you use same class structure implemented in paymentsreporting ingestion flow?)

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
```diff
@@ -26,50 +28,87 @@
...
 
-      List<Path> ingestionFlowFiles = retrieveFiles(ingestionFlowFileDTO);
+        if (ingestionFlowFiles != null && !ingestionFlowFiles.isEmpty()) {
+            ingestionFlowFiles.forEach(path -> {
+                File ingestionFlowFile = path.toFile();
+                log.debug("file from zip archive with name {} loaded successfully ", ingestionFlowFile.getName());
 
-     log.debug("Successfully retrieved the following files related to the ingestionFlowFileId {}: {}", ingestionFlowFileId, ingestionFlowFiles);
+                success.set(parseData(ingestionFlowFile));
```
antonioT90, line 68: you should collect iuf list, refactor this stream pipeline to map each file processing in order to build the resulting list (parseData should not return boolean, but the list of iuf.
You should do something like:

ingestionFlowFiles.stream()
   .map(this::parseFile) // log and File conversion, will be performed inside this method
   .flatMap(List::stream)
   .toList()

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
```diff
@@ -26,50 +28,87 @@
...
 
-  public TreasuryOpiIngestionActivityImpl(IngestionFlowFileDao ingestionFlowFileDao,
-                                          IngestionFlowFileRetrieverService ingestionFlowFileRetrieverService) {
-    this.ingestionflowFileType = IngestionFlowFileType.OPI;
-    this.ingestionFlowFileDao = ingestionFlowFileDao;
-    this.ingestionFlowFileRetrieverService = ingestionFlowFileRetrieverService;
-  }
+    @Override
+    public TreasuryIufResult processFile(Long ingestionFlowFileId) {
+        List<String> iufIuvList = new ArrayList<>();
```
antonioT90, line 50: is this just a iuf list?
if build using stream API we don't need to initialize it
```suggestion
        List<String> iufList ;
```

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
```diff
@@ -26,50 +28,87 @@
...
-        throw new IllegalArgumentException("invalid ingestionFlow file type " + ingestionFlowFileDTO.getFlowFileType());
-      }
-    return ingestionFlowFileDTO;
-  }
+        return ingestionFlowFileRetrieverService
+                .retrieveAndUnzipFile(Path.of(ingestionFlowFileDTO.getFilePath()), ingestionFlowFileDTO.getFileName());
+    }
 
-  private  List<Path> retrieveFiles(IngestionFlowFileDTO ingestionFlowFileDTO) throws IOException {
+    private boolean parseData(File ingestionFlowFile) {
```
antonioT90, line 90: if you plan read the content and return iuf list in an other task, let's just return an empty list

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
```diff
@@ -26,50 +28,87 @@
...
+
+
+        it.gov.pagopa.payhub.activities.xsd.treasury.opi14.FlussoGiornaleDiCassa flussoGiornaleDiCassa14 = null;
+        it.gov.pagopa.payhub.activities.xsd.treasury.opi161.FlussoGiornaleDiCassa flussoGiornaleDiCassa161 = null;
+
+        try {
```
antonioT90, line 96: there is not any way to know what should be right unmarshaller instead of try/catching?

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
domenicogi: Yes, clean operations are in another task. In this one there is the only opi validation

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
domenicogi: it was an additional control for test. Condition removed

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
domenicogi: moved code inside try/catch

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
domenicogi: refactor done

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
domenicogi: refactor done

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
domenicogi: It would be possible but other dependencies like DOM (Document Object Model) or XPath are required

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
```diff
@@ -26,50 +28,84 @@
...
+    private final IngestionFlowFileDao ingestionFlowFileDao;
+    private final IngestionFlowFileRetrieverService ingestionFlowFileRetrieverService;
+    private final TreasuryUnmarshallerService treasuryUnmarshallerService;
+
+
+    public TreasuryOpiIngestionActivityImpl(
+                                            IngestionFlowFileDao ingestionFlowFileDao,
+                                            IngestionFlowFileRetrieverService ingestionFlowFileRetrieverService,
+                                            TreasuryUnmarshallerService treasuryUnmarshallerService) {
+        this.ingestionflowFileType = IngestionFlowFileType.OPI;
```
antonioT90, line 41: this could be a constant of this class instead of instantiate it on the Constructor

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
```diff
@@ -26,50 +28,84 @@
...
+        }
+        if (flussoGiornaleDiCassa161 == null) {
+            try {
+                flussoGiornaleDiCassa14 = treasuryUnmarshallerService.unmarshalOpi14(ingestionFlowFile);
+                log.debug("file flussoGiornaleDiCassa with Id {} parsed successfully ", flussoGiornaleDiCassa14.getId());
+            } catch (Exception e) {
+                log.error("file flussoGiornaleDiCassa parsing error with opi 1.4 format {} ", e.getMessage());
+            }
+        }
+        return List.of();
```
antonioT90, line 108: can you add a TODO here with the id of the task to describe why we are returning an empty list?

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
```diff
@@ -26,50 +28,84 @@
...
+            log.error("file flussoGiornaleDiCassa parsing error with opi 1.6.1 format {} ", e.getMessage());
+        }
+        if (flussoGiornaleDiCassa161 == null) {
+            try {
+                flussoGiornaleDiCassa14 = treasuryUnmarshallerService.unmarshalOpi14(ingestionFlowFile);
+                log.debug("file flussoGiornaleDiCassa with Id {} parsed successfully ", flussoGiornaleDiCassa14.getId());
+            } catch (Exception e) {
+                log.error("file flussoGiornaleDiCassa parsing error with opi 1.4 format {} ", e.getMessage());
+            }
+        }
```
antonioT90, line 107: if we cannot understand the right model without opening the file, it could be ok this approach, but the first exception could not be an error, moreover, the second one yes, it's an error and we should throw an exception: let's introduce a custom Exception to be throw here (eg `TreasuryOpiInvalidFileException`)
```suggestion
            log.info("file flussoGiornaleDiCassa parsing error with opi 1.6.1 format {} ", e.getMessage());
            try {
                flussoGiornaleDiCassa14 = treasuryUnmarshallerService.unmarshalOpi14(ingestionFlowFile);
                log.debug("file flussoGiornaleDiCassa with Id {} parsed successfully ", flussoGiornaleDiCassa14.getId());
            } catch (Exception e) {
                log.info("file flussoGiornaleDiCassa parsing error with opi 1.4 format {} ", e.getMessage());
                throw new TreasuryOpiInvalidFileException("Cannot parse treasury Opi file " + ingestionFlowFile);
            }
        }
```

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
domenicogi: added a TODO message

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
domenicogi: removed association in costructor and aligned code

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
```diff
@@ -26,50 +27,82 @@
...
-        throw new IllegalArgumentException("invalid ingestionFlow file type " + ingestionFlowFileDTO.getFlowFileType());
-      }
-    return ingestionFlowFileDTO;
-  }
+        return ingestionFlowFileRetrieverService
+                .retrieveAndUnzipFile(Path.of(ingestionFlowFileDTO.getFilePath()), ingestionFlowFileDTO.getFileName());
+    }
 
-  private  List<Path> retrieveFiles(IngestionFlowFileDTO ingestionFlowFileDTO) throws IOException {
+    private List<String> parseData(Path ingestionFlowFilePath) {
```
antonioT90, line 85: Given the size, and probably it could grow due to the following tasks, probably it could be better if we create a new TrasuryOpiParserService

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/activity/treasury/TreasuryOpiIngestionActivityImpl.java
domenicogi: extract methon in a new service class

antonioT90 (approved):

PR #97, opened 2025-01-28
feat: P4ADEV-1981-WF-TransferClassification-service-implementation-for-CRUD-CRUD-invocation-from-activities

macacia:

- Pre-Deploy Test
    - [x] Unit
    - [ ] Integration (Narrow)
- Post-Deploy Test
    - [ ] Isolated Microservice
    - [ ] Broader Integration
    - [ ] Acceptance
    - [ ] Performance & Load

#### Types of changes

#### Checklist:

- [x] I have updated the documentation accordingly.


src/main/java/it/gov/pagopa/payhub/activities/connector/transfer/config/TransferApisHolder.java
```diff
@@ -0,0 +1,53 @@
+package it.gov.pagopa.payhub.activities.connector.transfer.config;
```
antonioT90, line 1: under connector package we should have one package per application not one package per entity!
You have to use debtpositions connector, using the same ApisHolder

antonioT90 (changes requested):

antonioT90 (approved):

PR #132, opened 2025-02-19
feat: P4ADEV-2195 Create activity process receipt

serdimic:
#### Description

Create activity process receipt
#### List of Changes

Create activity process receipt
#### Motivation and Context

Create activity process receipt

- Pre-Deploy Test
    - [X] Unit
    - [ ] Integration (Narrow)
- Post-Deploy Test
    - [ ] Isolated Microservice
    - [ ] Broader Integration
    - [ ] Acceptance
    - [ ] Performance & Load

#### Types of changes

#### Checklist:


src/main/java/it/gov/pagopa/payhub/activities/activity/ingestionflow/receipt/ReceiptPagopaIngestionActivityImpl.java
```diff
@@ -0,0 +1,97 @@
...
+
+  @Override
+  protected ReceiptPagopaIngestionFlowFileResult handleRetrievedFiles(List<Path> retrievedFiles, IngestionFlowFile ingestionFlowFileDTO) {
+    int ingestionFlowFilesRetrievedSize = retrievedFiles.size();
+    if (ingestionFlowFilesRetrievedSize != 1) {
+      return new ReceiptPagopaIngestionFlowFileResult(
+        null,
+        "Expected 1 file [" + ingestionFlowFileDTO.getFileName() + "], found " + ingestionFlowFilesRetrievedSize,
+        ingestionFlowFileDTO.getFileName()
+      );
```
antonioT90, line 64: don't do this, let's handle the error conditions throwing an exception using the message to fill the error_description.
Let's return the data required for the happy path.

ps: the discardFileName is the name of an optional file produced with the errors occurred during row elaboration! not the original file name

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/activity/ingestionflow/receipt/ReceiptPagopaIngestionActivityImpl.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+    Path fileToProcess = retrievedFiles.getFirst();
+
+    try {
+      //parse receipt file
+      PaSendRTV2Request paSendRTV2Request = receiptParserService.parseReceiptPagopaFile(fileToProcess, ingestionFlowFileDTO);
+
+      //map to DTO
+      ReceiptWithAdditionalNodeDataDTO receiptWithAdditionalNodeDataDTO = receiptMapper.map(paSendRTV2Request);
```
antonioT90, line 73: in order to keep simple main method (paSendRtV2 has not other usage than its mapping) let's move these rows on a separate `parseData` private method

src/main/java/it/gov/pagopa/payhub/activities/util/ConversionUtils.java
antonioT90, line 1: There is already a `it.gov.pagopa.payhub.activities.util.Utilities`, inside which there is a ZONEID Constant.
Can you move other conversion utilities implemented there (also the ZONEID constant) here?

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/activity/ingestionflow/receipt/ReceiptPagopaIngestionActivityImpl.java
serdimic: done

src/main/java/it/gov/pagopa/payhub/activities/util/ConversionUtils.java
serdimic: done

src/main/java/it/gov/pagopa/payhub/activities/activity/ingestionflow/receipt/ReceiptPagopaIngestionActivityImpl.java
```diff
@@ -0,0 +1,94 @@
...
+      receiptWithAdditionalNodeDataDTO.setReceiptId(receiptDTO.getReceiptId());
+
+      return new ReceiptPagopaIngestionFlowFileResult(receiptWithAdditionalNodeDataDTO);
+    } catch (Exception e) {
+      log.error("Unexpected error when processing receipt pagopa file[{}] with id[{}]",
+        fileToProcess, ingestionFlowFileDTO.getIngestionFlowFileId(), e);
+      throw new ReceiptIngestionFlowProcessingException(
+        "Unexpected error when processing receipt pagopa file with id[" + ingestionFlowFileDTO.getIngestionFlowFileId() + "]: " + e.getMessage(),
+        e);
+    }
```
antonioT90, line 83: If there are not checked exception to catch, let's avoid to build try/catch to wrap the unchecked exception into RetryableException.
Actually you are not adding any new info

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/activity/ingestionflow/receipt/ReceiptPagopaIngestionActivityImpl.java
serdimic: done

antonioT90 (approved):

PR #133, opened 2025-02-19
feat: P4ADEV-2126-activity-installment-ingestion-flow-file

LarissaASLeite:

- Pre-Deploy Test
    - [x] Unit
    - [ ] Integration (Narrow)
- Post-Deploy Test
    - [ ] Isolated Microservice
    - [ ] Broader Integration
    - [ ] Acceptance
    - [ ] Performance & Load

#### Types of changes

#### Checklist:


src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/debtposition/InstallmentProcessingService.java
```diff
@@ -0,0 +1,162 @@
...
+    /**
+     * Waits for a workflow to reach a terminal status.
+     *
+     * @param workflowId The ID of the workflow to monitor.
+     * @param installment The installment ingestion flow file DTO associated with the workflow.
+     * @param fileName The name of the file being processed.
+     * @param errorList The list where errors encountered during processing will be recorded.
+     * @return {@code true} if the workflow completed successfully, {@code false} if it terminated with an error or exceeded the retry limit.
+     */
+    private boolean waitForWorkflowCompletion(String workflowId, InstallmentIngestionFlowFileDTO installment,
```
antonioT90, line 116: move to separate class

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/treasury/TreasuryErrorsArchiverService.java
```diff
@@ -1,114 +1,32 @@
...
-            } else {
-                return null;
-            }
-        } catch (IOException e){
-            log.error("Something gone wrong while trying to archive error file!", e);
-            return null;
-        }
+    @Override
+    public void writeErrors(Path workingDirectory, IngestionFlowFile ingestionFlowFileDTO, List<TreasuryErrorDTO> errorList) {
+        List<String> headers = List.of("FileName", "Anno Bolletta", "Codice Bolletta", "Error Code", "Error Message");
```
antonioT90, line 28: set as class constant

src/main/java/it/gov/pagopa/payhub/activities/util/CsvOffsetDateTimeConverter.java
```diff
@@ -0,0 +1,18 @@
+package it.gov.pagopa.payhub.activities.util;
```
antonioT90, line 1:
```suggestion
package it.gov.pagopa.payhub.activities.util.csv;
```

src/main/java/it/gov/pagopa/payhub/activities/util/CsvOffsetDateTimeConverter.java
```diff
@@ -0,0 +1,18 @@
...
+import java.time.ZoneOffset;
+
+public class CsvOffsetDateTimeConverter extends AbstractBeanField<OffsetDateTime, String> {
+
+    @Override
+    protected OffsetDateTime convert(String value) {
+        if (value == null || value.isEmpty()) {
+            return null;
+        }
+        return LocalDate.parse(value).atStartOfDay().atOffset(ZoneOffset.UTC);
```
antonioT90, line 16: why should you read a LocalDate and convert it into OffsetDateTime at midnight UTC?

src/main/java/it/gov/pagopa/payhub/activities/mapper/ingestionflow/debtposition/InstallmentSynchronizeMapper.java
```diff
@@ -0,0 +1,103 @@
...
+                .balance(installmentIngestionFlowFileDTO.getBalance())
+                .flagMultibeneficiary(installmentIngestionFlowFileDTO.getFlagMultiBeneficiary())
+                .numberBeneficiary(installmentIngestionFlowFileDTO.getNumberBeneficiary() != null ? Long.valueOf(installmentIngestionFlowFileDTO.getNumberBeneficiary()) : null)
+                .transfersList(buildTransferList(installmentIngestionFlowFileDTO))
+                .build();
+    }
+
+    private List<TransferSynchronizeDTO> buildTransferList(InstallmentIngestionFlowFileDTO dto) {
+        if (Boolean.TRUE.equals(dto.getFlagMultiBeneficiary()) && dto.getNumberBeneficiary() != null && dto.getNumberBeneficiary() >= 2) {
+            return IntStream.rangeClosed(2, dto.getNumberBeneficiary())
```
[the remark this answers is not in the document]
Benedetta-fabbri: Because from the ingestionFile we don't have esplicit transfer with index 1, but I have to extrapolate from other data

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/debtposition/InstallmentProcessingService.java
```diff
@@ -0,0 +1,162 @@
...
+        long processedRows = installmentIngestionFlowFileDTOStream
+                .map(installmentIngestionFlowFileDTO -> {
+                    InstallmentSynchronizeDTO installmentSynchronizeDTO = installmentSynchronizeMapper.map(
+                            installmentIngestionFlowFileDTO,
+                            ingestionFlowFile.getIngestionFlowFileId(),
+                            ingestionFlowFile.getOrganizationId()
+                    );
+                    try {
+                        // see massive
+                        String workflowId = debtPositionService.installmentSynchronize(installmentSynchronizeDTO, false);
```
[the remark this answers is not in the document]
LarissaASLeite: @Benedetta-fabbri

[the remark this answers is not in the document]
Benedetta-fabbri: For the moment we have decided to call the GPD api for a single DP because their development for the massive v2.0 is not yet ready

src/main/java/it/gov/pagopa/payhub/activities/util/CsvOffsetDateTimeConverter.java
LarissaASLeite: The input format in our CSV files contains only dates (YYYY-MM-DD) without timezone information. Since OffsetDateTime requires a time and an offset, I assume midnight UTC as the default, since could not matter.

src/main/resources/config/application.yml
```diff
@@ -78,6 +78,9 @@ rest:
 
   process-executions:
     base-url: "\${PROCESS_EXECUTIONS_BASE_URL:http://localhost:8080}"
+
+  workflow-hub:
+    base-url: "\${WORKFLOW_HUB_BASE_URL:http://localhost:8080}"
     max-attempts: "\${PROCESS_EXECUTIONS_MAX_ATTEMPTS:3}" # It's counting also the first attempt
     wait-time-millis: "\${PROCESS_EXECUTIONS_WAIT_TIME_MILLIS:500}"
     print-body-when-error: "\${PROCESS_EXECUTIONS_PRINT_BODY_WHEN_ERROR:true}"
```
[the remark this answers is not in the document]
LarissaASLeite: https://github.com/pagopa/p4pa-payhub-deploy-aks/pull/50

src/main/java/it/gov/pagopa/payhub/activities/dto/ingestion/constants/CsvHeaders.java
antonioT90, line 1: don't create a constant class to store headers! They will not have other usage than the class that actually are using them! move their definition in the respective class and drop this class

src/main/java/it/gov/pagopa/payhub/activities/dto/ingestion/constants/WorkflowStatus.java
antonioT90, line 1: can we use Temporal enum instead?
see: it/gov/pagopa/pu/workflow/TemporalSpringBootIntegrationTest.java

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/ErrorArchiverService.java
```diff
@@ -0,0 +1,119 @@
...
+     *
+     * @param workingDirectory  The directory where the file should be created.
+     * @param ingestionFlowFile The metadata of the ingestion file.
+     * @param errorList         The list of errors to write.
+     * @param headers           The headers of the CSV file.
+     */
+    public void writeErrors(Path workingDirectory, IngestionFlowFile ingestionFlowFile,
+                            List<T> errorList, List<String> headers) {
+
+        if(errorList.isEmpty()){
```
antonioT90, line 51: Maybe it could be null?
```suggestion
        if(CollectionUtils.isEmpty(errorList)){
```

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/ErrorArchiverService.java
```diff
@@ -0,0 +1,119 @@
...
+                        .filter(f -> f.getFileName().toString().startsWith(ERRORFILE_PREFIX))
+                        .toList();
+            }
+
+            if (!errorFiles.isEmpty()) {
+
+                Path targetDirectory = createTargetDirectory(ingestionFlowFileDTO);
+
+                String zipFileName = ERRORFILE_PREFIX + Utilities.replaceFileExtension(ingestionFlowFileDTO.getFileName(), ".zip");
+                Path zipFile = Path.of(workingDirectory+"/"+zipFileName);
```
antonioT90, line 97:
```suggestion
                Path zipFile = Path.of(workingDirectory).resolve(zipFileName);
```

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/debtposition/WorkflowCompletionService.java
```diff
@@ -0,0 +1,89 @@
...
+@Service
+@Lazy
+@Slf4j
+public class WorkflowCompletionService {
+
+    private final WorkflowHubService workflowHubService;
+    private final int maxRetries;
+    private final int retryDelayMs;
+
+    public static final Set<String> TERMINAL_STATUSES = Set.of(COMPLETED, FAILED, CANCELLED, TERMINATED, TIMED_OUT);
```
antonioT90, line 26: please, link Temporal doc:
https://docs.temporal.io/workflows#status

src/main/java/it/gov/pagopa/payhub/activities/dto/debtposition/InstallmentIngestionFlowFileDTO.java
```diff
@@ -0,0 +1,132 @@
...
+    private String balance;
+
+    @CsvBindByName(column = "flagMultiBeneficiary")
+    private Boolean flagMultiBeneficiary;
+
+    @CsvBindByName(column = "numberBeneficiary")
+    private Integer numberBeneficiary;
+
+    @CsvBindAndJoinByName(column = "*_*", elementType = String.class)
+    private MultiValuedMap<String, String> transfers;
```
antonioT90, line 123: what do you think to formalize explicitly the number of possible transfers?

an approach *_* could be dangerouse

```suggestion
    @CsvBindAndJoinByName(column = "*_2", elementType = String.class)
    private MultiValuedMap<String, String> transfer2;
    
    @CsvBindAndJoinByName(column = "*_3", elementType = String.class)
    private MultiValuedMap<String, String> transfer3;
    
    @CsvBindAndJoinByName(column = "*_4", elementType = String.class)
    private MultiValuedMap<String, String> transfer4;
    
    @CsvBindAndJoinByName(column = "*_5", elementType = String.class)
    private MultiValuedMap<String, String> transfer5;
```

src/main/java/it/gov/pagopa/payhub/activities/dto/ingestion/CsvReadResult.java
```diff
@@ -0,0 +1,19 @@
...
+import lombok.Data;
+
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.Stream;
+
+@Data
+@AllArgsConstructor
+public class CsvReadResult<T> {
+    private final Stream<T> dataStream;
+    private final AtomicLong totalRows;
```
antonioT90, line 13: I see not advantage on handling the count on this manner...
totalRows will be not completed untile the stream is not totally read...

a structure like this could let think to be able to have the totalRows without even read the stream...
let's handle the counter directly on stream reading... (ps: why don't we just return the iterator itself?

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/dto/ingestion/constants/CsvHeaders.java
LarissaASLeite: sorry, i've read it wrong

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/ErrorArchiverService.java
LarissaASLeite: @domenicogi

build.gradle.kts
```diff
@@ -268,10 +268,42 @@ tasks.register("dependenciesBuild") {
...
+		"openApiGenerateWORKFLOWHUB"
 	)
 }
 
+tasks.register<GenerateTask>("openApiGenerateWORKFLOWHUB") {
+	group = "AutomaticallyGeneratedCode"
+	description = "openapi"
+
+	generatorName.set("java")
+	remoteInputSpec.set("https://raw.githubusercontent.com/pagopa/p4pa-workflow-hub/refs/heads/develop/openapi/generated.openapi.json")
```
antonioT90, line 281: don't use the generated, otherwise it will still create the ineherithed openApi

src/main/java/it/gov/pagopa/payhub/activities/mapper/ingestionflow/debtposition/InstallmentSynchronizeMapper.java
```diff
@@ -0,0 +1,109 @@
...
+                .amount(installmentIngestionFlowFileDTO.getAmount())
+                .debtPositionTypeCode(installmentIngestionFlowFileDTO.getDebtPositionTypeCode())
+                .paymentTypeCode(installmentIngestionFlowFileDTO.getPaymentTypeCode())
+                .remittanceInformation(installmentIngestionFlowFileDTO.getRemittanceInformation())
+                .legacyPaymentMetadata(installmentIngestionFlowFileDTO.getLegacyPaymentMetadata())
+                .flagPagoPaPayment(installmentIngestionFlowFileDTO.getFlagPagoPaPayment())
+                .balance(installmentIngestionFlowFileDTO.getBalance())
+                .flagMultibeneficiary(installmentIngestionFlowFileDTO.getFlagMultiBeneficiary())
+                .numberBeneficiary(installmentIngestionFlowFileDTO.getNumberBeneficiary() != null ? Long.valueOf(installmentIngestionFlowFileDTO.getNumberBeneficiary()) : null)
+                .transfersList(buildTransferList(installmentIngestionFlowFileDTO))
```
antonioT90, line 57: this is not the complete transferList?
Why don't we call it `additionalTransfers`?

src/main/java/it/gov/pagopa/payhub/activities/activity/ingestionflow/debtposition/InstallmentIngestionFlowFileActivityImpl.java
```diff
@@ -2,46 +2,81 @@
...
+        log.info("Processing file: {}", filePath);
+
+        try {
+            Iterator<InstallmentIngestionFlowFileDTO> csvIterator =
+                    csvService.readCsv(filePath, InstallmentIngestionFlowFileDTO.class, InstallmentIngestionFlowFileDTO::setIngestionFlowFileLineNumber);
+
+
+            Stream<InstallmentIngestionFlowFileDTO> stream = StreamSupport.stream(
+                    Spliterators.spliteratorUnknownSize(csvIterator, Spliterator.ORDERED), false
+            );
```
antonioT90, line 74: if you like to use Stream instead of Iterator, why don't you return directly the Stream (moving the creation code from here to CsvService) instead of returning an Iterator which you will not use?

Another point, you are transforming the iterator into Stream because you find simpler working with Stream?
ps: I have to read the entire PR before to understand if it's an advantage to transform the Iterator into Stream or not, so I'll let to you to think about it

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/debtposition/InstallmentProcessingService.java
```diff
@@ -0,0 +1,106 @@
...
+     * @param installmentIngestionFlowFileDTOStream Stream of installment ingestion flow file DTOs to be processed.
+     * @param ingestionFlowFile Metadata of the ingestion file containing details about the ingestion process.
+     * @param workingDirectory The directory where error files will be written if processing fails.
+     * @return An {@link InstallmentIngestionFlowFileResult} containing details about the processed rows, errors, and archived files.
+     */
+    public InstallmentIngestionFlowFileResult processInstallments(Stream<InstallmentIngestionFlowFileDTO> installmentIngestionFlowFileDTOStream,
+                                                                  IngestionFlowFile ingestionFlowFile,
+                                                                  Path workingDirectory) {
+        List<InstallmentErrorDTO> errorList = new ArrayList<>();
+        AtomicLong processedRows = new AtomicLong(0);
```
antonioT90, line 52: don't use Atomic* types if you have not to handle with concurrency! Here you are not working in multi-thread!
Just use long[] processedRows = {0L};

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/debtposition/WorkflowCompletionService.java
```diff
@@ -0,0 +1,110 @@
...
+    private final Set<WorkflowExecutionStatus> TERMINAL_STATUSES = Set.of(
+            WORKFLOW_EXECUTION_STATUS_FAILED,
+            WORKFLOW_EXECUTION_STATUS_TERMINATED,
+            WORKFLOW_EXECUTION_STATUS_CANCELED,
+            WORKFLOW_EXECUTION_STATUS_TIMED_OUT,
+            WORKFLOW_EXECUTION_STATUS_COMPLETED
+    );
+
+    public WorkflowCompletionService(WorkflowHubService workflowHubService,
+                                     @Value("${ingestion-flow-files.dp-installments.wf-await.max-waiting-minutes:5}") double maxWaitingMinutes,
```
antonioT90, line 36: int or long, why double?
```suggestion
                                     @Value("${ingestion-flow-files.dp-installments.wf-await.max-waiting-minutes:5}") int maxWaitingMinutes,
```

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/debtposition/InstallmentProcessingService.java
```diff
@@ -0,0 +1,106 @@
...
+
+        long totalRows = installmentIngestionFlowFileDTOStream
```
antonioT90, line 54: this is not totalRows if you are applying a filter... this is the number of elements which have passed the filter.
You could use the filter to count just the correctly processedRows, while inside the filter, as first thing, you could increment the totalRows, handled as an explicit counter (instead of processedRows)

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/ErrorArchiverService.java
```diff
@@ -0,0 +1,120 @@
...
+            } else {
+                return null;
+            }
+        } catch (IOException e){
+            log.error("Something gone wrong while trying to archive error file!", e);
+            return null;
+        }
+    }
+
+    public Path createTargetDirectory(IngestionFlowFile ingestionFlowFileDTO) {
```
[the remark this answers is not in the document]
LarissaASLeite: this method is no longer needed once i deleted the field 

`/** Discarded file name path */`
 `private String discardedFilePath;`

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/debtposition/WorkflowCompletionService.java
LarissaASLeite: I used double for maxWaitingMinutes because, with int, the formula `this.maxRetries = (int) ((maxWaitingMinutes * 60_000) / retryDelayMs);` would make maxWaitingMinutes round to 1 as the minimum value. This would lead to too many retries and longer test execution times.

I could switch to long maxWaitingMillis instead of int maxWaitingMinutes. This way, we could setvalues like 300ms for quick tests.

What do you think?

src/main/java/it/gov/pagopa/payhub/activities/activity/ingestionflow/debtposition/InstallmentIngestionFlowFileActivityImpl.java
LarissaASLeite: Actually is the same for me, i do find Stream more easy to work with, more readable and with more functionality and i was trying to reuse the code as it was, that is to say conserve the processInstallments() method with the Stream, on the other hand iterator maybe can offer more details about the single row. I'm pushing an iterator version for you

src/main/java/it/gov/pagopa/payhub/activities/mapper/ingestionflow/debtposition/InstallmentSynchronizeMapper.java
Benedetta-fabbri: I have to create the first transfer based on the installment data

Benedetta-fabbri: I'll change the openapi

src/main/java/it/gov/pagopa/payhub/activities/util/CsvOffsetDateTimeConverter.java
LarissaASLeite: Installment save the fields 

`private OffsetDateTime notificationDate;`
`private OffsetDateTime validityDate;`
`private OffsetDateTime dueDate;`

if i don't convert while reading the file, i need the converter anyways to map to `InstallmentSynchronizeDTO`

antonioT90: Is it right that these fields are OffsetDateTime and not LocalDate?

CC: @robertopasetti, @serdimic, @Benedetta-fabbri

src/main/java/it/gov/pagopa/payhub/activities/mapper/ingestionflow/debtposition/InstallmentSynchronizeMapper.java
```diff
@@ -0,0 +1,107 @@
...
+        MultiValuedMap<String, String> transferMap = getTransferMapByIndex(dto, index);
+
+        if (transferMap == null) {
+            throw new IllegalStateException("Missing or empty transfer map for index: " + index);
+        }
+
+        return TransferSynchronizeDTO.builder()
+                .orgFiscalCode(getFirstValue(transferMap, "orgFiscalCode"))
+                .orgName(getFirstValue(transferMap, "orgName"))
+                .amount(new BigDecimal(getFirstValue(transferMap, "amount")))
```
antonioT90, line 83: what about to convert amount into cents and send to debt-position API already as Cents?

@Benedetta-fabbri what do you think about it?

src/main/java/it/gov/pagopa/payhub/activities/mapper/ingestionflow/debtposition/InstallmentSynchronizeMapper.java
```diff
@@ -0,0 +1,107 @@
...
+                .fullName(installmentIngestionFlowFileDTO.getFullName())
+                .address(installmentIngestionFlowFileDTO.getAddress())
+                .civic(installmentIngestionFlowFileDTO.getCivic())
+                .postalCode(installmentIngestionFlowFileDTO.getPostalCode())
+                .location(installmentIngestionFlowFileDTO.getLocation())
+                .province(installmentIngestionFlowFileDTO.getProvince())
+                .nation(installmentIngestionFlowFileDTO.getNation())
+                .email(installmentIngestionFlowFileDTO.getEmail())
+                .dueDate(installmentIngestionFlowFileDTO.getDueDate())
+                .amount(installmentIngestionFlowFileDTO.getAmount())
```
antonioT90, line 49: what about to convert amount into cents and send to debt-position API already as Cents?

@Benedetta-fabbri what do you think about it?

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/debtposition/WorkflowCompletionService.java
```diff
@@ -0,0 +1,110 @@
...
+
+import java.util.List;
+import java.util.Set;
+
+import static io.temporal.api.enums.v1.WorkflowExecutionStatus.*;
+
+@Service
+@Lazy
+@Slf4j
+public class WorkflowCompletionService {
```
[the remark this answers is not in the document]
antonioT90: move it also on the suggested package!

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/debtposition/WorkflowCompletionService.java
antonioT90: you have to cast it during the calculation of the maxRetries, not expect it to be already a double!

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/debtposition/InstallmentProcessingService.java
```diff
@@ -0,0 +1,101 @@
...
+                String workflowId = debtPositionService.installmentSynchronize("ORDINARY_SIL", installmentSynchronizeDTO, false);
+                if (workflowId != null) {
+                    boolean workflowCompleted = dpInstallmentsWorkflowCompletionService.waitForWorkflowCompletion(
+                            workflowId, installment, ingestionFlowFile.getFileName(), errorList
+                    );
+                    log.info("Workflow with id {} completed", workflowId);
+                    if (workflowCompleted) {
+                        processedRows++;
+                    }
+                } else {
```
antonioT90, line 74: move these rows into a separate `waitWorkflowCompletion` method, which will return true if the row has been processed (if worflowId was null or the workflow completed sucessfully).
The you will update processedRows once if the method returned true

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/exception/ingestionflow/TooManyAttemptsException.java
```diff
@@ -0,0 +1,10 @@
+package it.gov.pagopa.payhub.activities.exception.ingestionflow;
+
+import it.gov.pagopa.payhub.activities.exception.NotRetryableActivityException;
+
+public class TooManyAttemptsException extends NotRetryableActivityException {
```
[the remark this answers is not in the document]
LarissaASLeite: can i extend RuntimeException?

src/main/java/it/gov/pagopa/payhub/activities/util/CsvOffsetDateTimeConverter.java
serdimic: For what I remember, we agred to use OffsetDateTime whenever the corresponding field type on DB was "timestamp".
For DB fields of type "date" (i.e. without time and timezone), I'm not totally sure (can't find anything about this on Slack) but it seems to me we decided to use LocalDate.

In this particular case, I don't know what is the mapping of these fields on DB.

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/ErrorArchiverService.java
```diff
@@ -0,0 +1,112 @@
...
+     * Writes error data into a CSV file.
+     *
+     * @param workingDirectory  The directory where the file should be created.
+     * @param ingestionFlowFile The metadata of the ingestion file.
+     * @param errorList         The list of errors to write.
+     * @param headers           The headers of the CSV file.
+     */
+    public void writeErrors(Path workingDirectory, IngestionFlowFile ingestionFlowFile,
+                            List<T> errorList, List<String[]> headers) {
+
```
antonioT90, line 50: the headers could be retrieved with an abstract get method instead of adding it as a parameter.
this way the public method could be just once, and the only exposed by this base service

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/debtposition/InstallmentProcessingService.java
```diff
@@ -0,0 +1,101 @@
...
+                    processedRows++;
+                }
+            } catch (Exception e) {
+                log.error("Error processing installment {}: {}", installment.getIud(), e.getMessage());
+                InstallmentErrorDTO error = dpInstallmentsWorkflowCompletionService.buildInstallmentErrorDTO(
+                        ingestionFlowFile.getFileName(), installment, null, "PROCESS_EXCEPTION", e.getMessage());
```
antonioT90, line 80: what's the purpose of this method `dpInstallmentsWorkflowCompletionService.buildInstallmentErrorDTO` it's just calling the constructor/builder...
just call the constructor/builder here instead of exposing a method to be reused here...
Otherwise, move the method elsewhere: why should I expect to use a service called DpInstallmentsWorkflowCompletionService to create the ErrorDto?

ps: a methodFactory could have a sense if you have to hide some inner details of creating an object... otherwise let's just call the new

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/debtposition/InstallmentProcessingService.java
```diff
@@ -0,0 +1,101 @@
...
+                log.info("Current error list size after handleProcessingError: {}", errorList.size());
+            }
+        }
+
+        String errorsZipFileName = null;
+        if (!errorList.isEmpty()) {
+            installmentErrorsArchiverService.writeErrors(workingDirectory, ingestionFlowFile, errorList);
+            errorsZipFileName = installmentErrorsArchiverService.archiveErrorFiles(workingDirectory, ingestionFlowFile);
+            log.info("Error file archived at: {}", errorsZipFileName);
+        }
```
antonioT90, line 91: move into private `archiveErrorFiles`

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/debtposition/WorkflowCompletionService.java
```diff
@@ -0,0 +1,85 @@
...
+    public WorkflowExecutionStatus waitTerminationStatus(String workflowId, int maxRetries, int retryDelayMs) throws TooManyAttemptsException {
+
+        int attempts = 0;
+        WorkflowExecutionStatus workflowStatus;
+
+        do {
+            WorkflowStatusDTO statusDTO = workflowHubService.getWorkflowStatus(workflowId);
+            String status = statusDTO.getStatus();
+            workflowStatus = mapToString(status);
+            log.info("Workflow {} status: {}", workflowId, status);
```
antonioT90, line 54:
```suggestion
            WorkflowStatusDTO statusDTO = workflowHubService.getWorkflowStatus(workflowId);
             log.debug("Retrieved workflow status: {}, statusDTO);
            workflowStatus = transcodeStatus(statusDTO.getStatus());
```

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/util/CsvOffsetDateTimeConverter.java
robertopasetti: I think I can't add any contribution to this discussion, I do not have enough elements to choose between OffsetDateTime and LocalDate

antonioT90: @robertopasetti the doubt is functional:
are notificationDate, validityDate, dueDate just date or date/time fields?

According to the CSV that Larissa is reading, these are just dates, not date/time

@LarissaASLeite is transforming those data in date/time setting the midnight UTC (which is 1 or 2 hours behind Europe/Rome).

Maybe we have chose the wrong data type for these fields

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/debtposition/InstallmentProcessingService.java
```diff
@@ -0,0 +1,105 @@
...
+
+        while (iterator.hasNext()) {
+            totalRows++;
+
+            InstallmentIngestionFlowFileDTO installment = iterator.next();
+            InstallmentSynchronizeDTO installmentSynchronizeDTO = installmentSynchronizeMapper.map(
+                    installment,
+                    ingestionFlowFile.getIngestionFlowFileId(),
+                    ingestionFlowFile.getOrganizationId()
+            );
```
antonioT90, line 63: probably it cannot go into error, but let's move it inside the try/catch block ;)

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/debtposition/InstallmentProcessingService.java
```diff
@@ -0,0 +1,105 @@
...
+            try {
+
+                // For the moment we have decided to call the GPD api for a single DP because their development for the massive v2.0 is not yet ready
+                String workflowId = debtPositionService.installmentSynchronize(ORDINARY_SIL, installmentSynchronizeDTO, false);
```
antonioT90, line 67: now we can send massive true ;)

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/dto/ingestion/debtposition/InstallmentIngestionFlowFileDTO.java
```diff
@@ -0,0 +1,141 @@
...
+    private String nation;
+
+    @CsvBindByName(column = "email")
+    private String email;
+
+    @CsvCustomBindByName(column = "dueDate", converter = CsvOffsetDateTimeConverter.class)
+    private OffsetDateTime dueDate;
+
+    @CsvBindByName(column = "amount", required = true)
+    private Long amountCents;
```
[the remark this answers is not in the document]
robertopasetti: We agreed to keep columns header in italian but to manage them in english internally. did I miss something?

src/main/java/it/gov/pagopa/payhub/activities/util/CsvOffsetDateTimeConverter.java
robertopasetti: notificationDate->date/time, validityDate and dueDate date fields

src/main/java/it/gov/pagopa/payhub/activities/dto/ingestion/debtposition/InstallmentIngestionFlowFileDTO.java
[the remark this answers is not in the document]
antonioT90: Actually we are supporting only the new file version, which has column name in english.
When we will introduce the support to the old italian version, which is a subset of the actual columns, we will use the OpenCsv profile functionality to configure on the DTO an alternative CsvBindByName configuration with the italian name.

Do you agree @LarissaASLeite?

src/main/java/it/gov/pagopa/payhub/activities/dto/ingestion/debtposition/InstallmentIngestionFlowFileDTO.java
```diff
@@ -0,0 +1,142 @@
...
+import java.time.OffsetDateTime;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+public class InstallmentIngestionFlowFileDTO {
+
+    @CsvIgnore
+    private Long ingestionFlowFileLineNumber;
```
antonioT90, line 24: This field is not filled... should we delete it?

src/main/java/it/gov/pagopa/payhub/activities/dto/ingestion/debtposition/InstallmentIngestionFlowFileDTO.java
```diff
@@ -0,0 +1,142 @@
...
+
+    @CsvBindByName(column = "azione", required = true)
+    private ActionEnum action;
```
antonioT90, line 27: let's introduce italian fields just for `legacy` version
```suggestion
    @CsvBindByNames({
       @CsvBindByName(column = "azione", required = true, profile = "legacy"),
       @CsvBindByName(column = "action", required = true)
    })
    private ActionEnum action;
```

src/main/java/it/gov/pagopa/payhub/activities/util/CsvOffsetDateTimeConverter.java
antonioT90: @LarissaASLeite the notificationDate inside the CSV is a date or date/time?

src/main/java/it/gov/pagopa/payhub/activities/mapper/ingestionflow/debtposition/InstallmentSynchronizeMapper.java
```diff
@@ -0,0 +1,109 @@
...
+@Service
+@Lazy
+public class InstallmentSynchronizeMapper {
+
+    public InstallmentSynchronizeDTO map(InstallmentIngestionFlowFileDTO installmentIngestionFlowFileDTO,
+                                         Long ingestionFlowFileId,
+                                         Long organizationId) {
+        return InstallmentSynchronizeDTO.builder()
+                .ingestionFlowFileId(ingestionFlowFileId)
+                .ingestionFlowFileLineNumber(installmentIngestionFlowFileDTO.getIngestionFlowFileLineNumber())
```
antonioT90, line 26: is this field really updated?

src/main/java/it/gov/pagopa/payhub/activities/connector/debtposition/client/DebtPositionClient.java
```diff
@@ -25,4 +27,12 @@ public DebtPositionDTO finalizeSyncStatus(String accessToken, Long debtPositionI
     public DebtPositionDTO checkAndUpdateInstallmentExpiration(String accessToken, Long debtPositionId){
         return debtPositionApisHolder.getDebtPositionApi(accessToken).checkAndUpdateInstallmentExpiration(debtPositionId);
     }
+
+    public String installmentSynchronize(String accessToken, String origin, InstallmentSynchronizeDTO installmentSynchronizeDTO, Boolean massive) {
```
antonioT90, line 31: Use the enum until client invoke, here, you will transcode into String
```suggestion
    public String installmentSynchronize(String accessToken, DebtPositionOrigin origin, InstallmentSynchronizeDTO installmentSynchronizeDTO, Boolean massive) {
```

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/util/CsvOffsetDateTimeConverter.java
antonioT90: @LarissaASLeite 
I changed dueDate and validityDate data type, you should see them as LocalDate now.
notificationDate will be handled as DateTime: id you read it as LocalDate, transform it into Midnight Europe/Rome (there should be a constant with the ZoneID already set)

LarissaASLeite: All dates from the file are ISO 8601 [YYYY]-[MM]-[DD], so just date

antonioT90: ok, let's convert into OffsetDateTime 23:59:59.999 (LocalTime.MAX) of Europe/Rome

ps: @robertopasetti the ACA API are expecting a date/time field for dueDate... i'have already changed it into LocalDate

robertopasetti: it should be the same for GPD

src/main/java/it/gov/pagopa/payhub/activities/dto/ingestion/debtposition/InstallmentIngestionFlowFileDTO.java
LarissaASLeite: I didn't use `@CsvBindByNames` because is deprecated, we just need the two `@CsvBindByName`

src/main/java/it/gov/pagopa/payhub/activities/mapper/ingestionflow/debtposition/InstallmentSynchronizeMapper.java
antonioT90: do we really need this field? could it be ambiguous to have it stored here?.
 Where actually you are setting it, you are invoking the code that should then use it, maybe it could be clearer to remove from this DTO and just pass the value to invoked method `debtPositionService.installmentSynchronize(...)`

src/main/java/it/gov/pagopa/payhub/activities/util/csv/CsvOffsetDateTimeConverter.java
antonioT90, line 1: Is this class still used?

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/util/csv/CsvOffsetDateTimeConverter.java
LarissaASLeite: no, just deleted

antonioT90 (approved):

PR #141, opened 2025-02-25
fix: P4ADEV-001 add method for total amount cents

domenicogi:

- Pre-Deploy Test
    - [X] Unit
    - [ ] Integration (Narrow)
- Post-Deploy Test
    - [ ] Isolated Microservice
    - [ ] Broader Integration
    - [ ] Acceptance
    - [ ] Performance & Load

#### Types of changes

#### Checklist:


src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/TransferClassifier.java
```diff
@@ -46,6 +46,16 @@ default Long getAmountCents(PaymentsReporting paymentsReportingDTO) {
 		return Optional.ofNullable(paymentsReportingDTO).map(PaymentsReporting::getAmountPaidCents).orElse(0L);
 	}
 
+	/**
+	 * Extracts the amount in cents from the payment reporting data.
+	 *
+	 * @param paymentsReportingDTO the payment reporting data.
+	 * @return the amount in cents, or {@code null} if the amount is not available.
+	 */
+	default Long getTotalAmountCents(PaymentsReporting paymentsReportingDTO) {
```
antonioT90, line 55: why have you add this method? maybe the error is that the method `Long getAmountCents(PaymentsReporting paymentsReportingDTO)` is reading the wrong field, so you have to fix it instead of add another method...

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/TransferClassifier.java
domenicogi: treasury must be compared with the total amount of the payment reporting unlike other cases where the comparison is between single amounts.

antonioT90: what do you think if we rename those methods into:

getTransferAmount and getInstallmentAmount?

antonioT90: These overloaded methods are intended to extract the amount to compare between Transfer, Treasury and PaymentsReporting, thus, if we have different type of amounts, we are interested into understand what king of amount we are talking about (or what kind of amount we expect it should be compared to).

Thus, yes, please, let's rename the orginal methods

CC: @marcomatteuccieng, @macacia

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/TransferClassifier.java
antonioT90, line 1: I don't agree on these names...
It has not value to configure on the method name, the name of the entity involved: it's already explicitly defined on the parameter itself!

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/TransferClassifier.java
```diff
@@ -32,7 +32,7 @@ public interface TransferClassifier {
 	 * @param transferDTO the transfer data.
 	 * @return the amount in cents, or {@code null} if the amount is not available.
 	 */
-	default Long getAmountCents(Transfer transferDTO) {
+	default Long getAmountCentsFromTransfer(Transfer transferDTO) {
```
antonioT90, line 35: This is explicit, we could let ASIS
```suggestion
	default Long getAmountCents(Transfer transferDTO) {
```

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/TransferClassifier.java
```diff
@@ -42,17 +42,27 @@ default Long getAmountCents(Transfer transferDTO) {
 	 * @param paymentsReportingDTO the payment reporting data.
 	 * @return the amount in cents, or {@code null} if the amount is not available.
 	 */
-	default Long getAmountCents(PaymentsReporting paymentsReportingDTO) {
+	default Long getAmountPaidCentsFromPaymentsReporting(PaymentsReporting paymentsReportingDTO) {
```
antonioT90, line 45: Why not the following?
```suggestion
	default Long getTransferAmountCents(PaymentsReporting paymentsReportingDTO) {
```

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/TransferClassifier.java
```diff
@@ -42,17 +42,27 @@ default Long getAmountCents(Transfer transferDTO) {
...
 		return Optional.ofNullable(paymentsReportingDTO).map(PaymentsReporting::getAmountPaidCents).orElse(0L);
 	}
 
+	/**
+	 * Extracts the total amount in cents from the payment reporting data.
+	 *
+	 * @param paymentsReportingDTO the payment reporting data.
+	 * @return the amount in cents, or {@code null} if the amount is not available.
+	 */
+	default Long getTotalAmountCentsFromPaymentsReporting(PaymentsReporting paymentsReportingDTO) {
```
antonioT90, line 55: why not the following?
```suggestion
	default Long getIufAmountCents(PaymentsReporting paymentsReportingDTO) {
```

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/TransferClassifier.java
```diff
@@ -42,17 +42,27 @@ default Long getAmountCents(Transfer transferDTO) {
...
+	}
+
 	/**
 	 * Extracts the amount in cents from the treasury data.
 	 *
 	 * @param treasury the treasury data.
 	 * @return the amount in cents, or {@code null} if the amount is not available.
 	 */
-	default Long getAmountCents(Treasury treasury) {
+	default Long getBillAmountCentsFromTreasury(Treasury treasury) {
```
antonioT90, line 65: why not the following?
```suggestion
	default Long getIufAmountCents(Treasury treasury) {
```

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/IufTesDivImpClassifier.java
```diff
@@ -13,7 +13,7 @@ public class IufTesDivImpClassifier implements TransferClassifier {
 
 	@Override
 	public ClassificationsEnum classify(Transfer transferDTO, PaymentsReporting paymentsReportingDTO, Treasury treasuryDTO) {
-		if (paymentsReportingDTO != null && treasuryDTO != null	&& !getAmountCents(paymentsReportingDTO).equals(treasuryDTO.getBillAmountCents())) {
+		if (paymentsReportingDTO != null && treasuryDTO != null	&& !getIufAmountCents(paymentsReportingDTO).equals(treasuryDTO.getBillAmountCents())) {
```
antonioT90, line 16:
```suggestion
		if (paymentsReportingDTO != null && treasuryDTO != null	&& !getIufAmountCents(paymentsReportingDTO).equals(getIufAmountCents(treasuryDTO))) {
```

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/RtIufTesClassifier.java
```diff
@@ -14,8 +14,8 @@ public class RtIufTesClassifier implements TransferClassifier {
 	@Override
 	public ClassificationsEnum classify(Transfer transferDTO, PaymentsReporting paymentsReportingDTO, Treasury treasuryDTO) {
 		if(transferDTO != null && paymentsReportingDTO != null && treasuryDTO != null &&
-			getAmountCents(transferDTO).equals(getAmountCents(paymentsReportingDTO)) &&
-			getAmountCents(transferDTO).equals(getAmountCents(treasuryDTO))) {
+			getAmountCents(transferDTO).equals(getTransferAmountCents(paymentsReportingDTO)) &&
+			getAmountCents(transferDTO).equals(getIufAmountCents(treasuryDTO))) {
```
antonioT90, line 18: how could this equal condition be possible? transferAmount equal to IufAmount?
@macacia @marcomatteuccieng

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/RtNoIufClassifier.java
```diff
@@ -16,10 +16,10 @@ public ClassificationsEnum classify(Transfer transferDTO, PaymentsReporting paym
 		if(transferDTO != null && paymentsReportingDTO == null && treasuryDTO == null) {
 			return ClassificationsEnum.RT_NO_IUF;
 		}
-		if (transferDTO != null && paymentsReportingDTO != null && treasuryDTO == null && !getAmountCents(transferDTO).equals(getAmountCents(paymentsReportingDTO))) {
+		if (transferDTO != null && paymentsReportingDTO != null && treasuryDTO == null && !getAmountCents(transferDTO).equals(getTransferAmountCents(paymentsReportingDTO))) {
 			return ClassificationsEnum.RT_NO_IUF;
 		}
-		if (transferDTO != null && treasuryDTO != null && paymentsReportingDTO == null && !getAmountCents(transferDTO).equals(getAmountCents(treasuryDTO))) {
+		if (transferDTO != null && treasuryDTO != null && paymentsReportingDTO == null && !getAmountCents(transferDTO).equals(getIufAmountCents(treasuryDTO))) {
```
antonioT90, line 22: how could this equal condition be possible? transferAmount equal to IufAmount?
@macacia @marcomatteuccieng

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/RtTesClassifier.java
```diff
@@ -14,7 +14,7 @@ public class RtTesClassifier implements TransferClassifier {
 	@Override
 	public ClassificationsEnum classify(Transfer transferDTO, PaymentsReporting paymentsReportingDTO, Treasury treasuryDTO) {
 		if(transferDTO != null && treasuryDTO != null && paymentsReportingDTO == null &&
-			getAmountCents(transferDTO).equals(getAmountCents(treasuryDTO))) {
+			getAmountCents(transferDTO).equals(getIufAmountCents(treasuryDTO))) {
```
antonioT90, line 17: how could this equal condition be possible? transferAmount equal to IufAmount?
@macacia @marcomatteuccieng

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/RtIufTesClassifier.java
marcomatteuccieng: In case of ingestion of treasury-iuv. This case is not handled today because it is not actually used by any organization but present in the global use case. In addition to this, there is a problem of uniqueness as the transfer index is missing.

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/RtNoIufClassifier.java
marcomatteuccieng: In case of ingestion of treasury-iuv. This case is not handled today because it is not actually used by any organization but present in the global use case. In addition to this, there is a problem of uniqueness as the transfer index is missing.

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/RtTesClassifier.java
marcomatteuccieng: In case of ingestion of treasury-iuv. This case is not handled today because it is not actually used by any organization but present in the global use case. In addition to this, there is a problem of uniqueness as the transfer index is missing.

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/RtIufTesClassifier.java
antonioT90: Thus, actually this label would never be applied ASIS, because iufAmounts will never be equal to transferAmount.

maybe we should simply fix changing the amount check using treasury in the following way (comparing paymentsReporting and treasury):
getIufAmountCents(paymentsReportingDTO).equals(getIufAmountCents(treasuryDTO)

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/RtNoIufClassifier.java
antonioT90: we want to let this label never applied if there is a treasury, or we want not to check the amount?

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/RtTesClassifier.java
antonioT90: we want to never this label be applied, or we want to remove the amount check?

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/RtIufTesClassifier.java
marcomatteuccieng: I agree

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/RtNoIufClassifier.java
marcomatteuccieng: do not check the amount

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/RtTesClassifier.java
marcomatteuccieng: this case seems correct to me, in the treasury-iuv case if the amounts coincide it is correct to assign the RT_TEST label.
So, the case is: if I find a treasury with IUV valued and the amount matches.

antonioT90: when we will introduce the TreasuryIuv, it will be a separate object, because the TreasuryIuv will be searched using a different query, so we will have separate object (between parameters we will have 2 treasury: treasuryIuf and treasuryIuv, bot of Treasury type).
we could already introduce it, in order to fix all the checks in order to consider it, but actually we will set it as null

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/RtNoIufClassifier.java
domenicogi: done

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/RtTesClassifier.java
domenicogi: leave as is

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/RtNoIufClassifier.java
antonioT90: or better, the particular condition when the treasury is present and the paymentReporting not. there are is a condition where the treasury is also null

antonioT90: The label RT_NO_IUF is intended if the amounts are not equal! @marcomatteuccieng this way we are applying the label always if there is the treasury and the transfer, but not the paymentsReporting.

This cannot happen, but removing the amount check is also an error: we should or remove the entire if, or introduce a parameter treasuryIuv, from which we will read the transferAmount and compare that with the transfer

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/RtTesClassifier.java
antonioT90: we cannot leave ASIS... This useCase actually cannot happen because this will be not possible: treasury!=null && paymentsReporting==null

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/IufNoTesClassifier.java
```diff
@@ -12,8 +13,8 @@
 public class IufNoTesClassifier implements TransferClassifier {
 
 	@Override
-	public ClassificationsEnum classify(Transfer transferDTO, PaymentsReporting paymentsReportingDTO, Treasury treasuryDTO) {
-		if(paymentsReportingDTO != null && treasuryDTO == null) {
+	public ClassificationsEnum classify(Transfer transferDTO, PaymentsReporting paymentsReportingDTO, TreasuryIuf treasuryIuf, TreasuryIuv treasuryIuv) {
+		if(paymentsReportingDTO != null && treasuryIuf == null && treasuryIuv == null) {
```
marcomatteuccieng, line 17: The control "if(paymentsReportingDTO != null && treasuryIuf == null )" is sufficient because in the treasuryIuv only the IUV is stored, not the IUF, but the implemented control has no impact on the result

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/RtNoIufClassifier.java
```diff
@@ -12,14 +13,14 @@
...
 			return ClassificationsEnum.RT_NO_IUF;
 		}
-		if (transferDTO != null && paymentsReportingDTO != null && treasuryDTO == null && !getAmountCents(transferDTO).equals(getAmountCents(paymentsReportingDTO))) {
+		if (transferDTO != null && paymentsReportingDTO != null && treasuryIuv == null && !getAmountCents(transferDTO).equals(getTransferAmountCents(paymentsReportingDTO))) {
 			return ClassificationsEnum.RT_NO_IUF;
 		}
-		if (transferDTO != null && treasuryDTO != null && paymentsReportingDTO == null && !getAmountCents(transferDTO).equals(getAmountCents(treasuryDTO))) {
+		if (transferDTO != null && treasuryIuv != null && paymentsReportingDTO == null && !getAmountCents(transferDTO).equals(getTransferAmountCents(treasuryIuv))) {
 			return ClassificationsEnum.RT_NO_IUF;
 		}
```
marcomatteuccieng, line 25: The only correct checks are: 
"if(transferDTO != null && paymentReportingDTO == null)
if (transferDTO != null && paymentReportingDTO != null && !getAmountCents(transferDTO).equals(getAmountCents(paymentsReportingDTO)))" 
Since only transfer and Payments_reporting are checked, treasury is out of scope for the RT_NO_IUF classification use case. The label indicates that Transfer is not reported by Payments_reporting or is reported but the amount does not match.

src/main/java/it/gov/pagopa/payhub/activities/service/classifications/trclassifiers/RtTesClassifier.java
```diff
@@ -12,9 +13,9 @@
 public class RtTesClassifier implements TransferClassifier {
 
 	@Override
-	public ClassificationsEnum classify(Transfer transferDTO, PaymentsReporting paymentsReportingDTO, Treasury treasuryDTO) {
-		if(transferDTO != null && treasuryDTO != null && paymentsReportingDTO == null &&
-			getAmountCents(transferDTO).equals(getAmountCents(treasuryDTO))) {
+	public ClassificationsEnum classify(Transfer transferDTO, PaymentsReporting paymentsReportingDTO, TreasuryIuf treasuryIuf, TreasuryIuv treasuryIuv) {
+		if(transferDTO != null && treasuryIuv != null && paymentsReportingDTO == null &&
+			getAmountCents(transferDTO).equals(getTransferAmountCents(treasuryIuv))) {
 			return ClassificationsEnum.RT_TES;
```
marcomatteuccieng, line 19: The correct check is 		
"if(transferDTO != null && treasuryIuv != null&&	getAmountCents(transferDTO).equals(getTransferAmountCents(treasuryIuv)))"
Because the payments_reporting can be present or not but this non impact in this classification.

antonioT90 (approved):

PR #154, opened 2025-03-06
fix: P4ADEV-2356-fix-notification-request-mapper

LarissaASLeite:

- Pre-Deploy Test
    - [x] Unit
    - [ ] Integration (Narrow)
- Post-Deploy Test
    - [ ] Isolated Microservice
    - [ ] Broader Integration
    - [ ] Acceptance
    - [ ] Performance & Load

#### Types of changes

- [x] PATCH - Bug fix (backwards compatible bug fixes)

#### Checklist:


src/main/java/it/gov/pagopa/payhub/activities/connector/ionotification/client/IoNotificationClient.java
```diff
@@ -15,8 +16,8 @@ public IoNotificationClient(IoNotificationApisHolder ioNotificationApisHolder) {
         this.ioNotificationApisHolder = ioNotificationApisHolder;
     }
 
-    public void sendMessage(NotificationRequestDTO notificationQueueDTO, String accessToken) {
-        ioNotificationApisHolder.getIoNotificationApi(accessToken)
+    public MessageResponseDTO sendMessage(NotificationRequestDTO notificationQueueDTO, String accessToken) {
```
antonioT90, line 19: could it be 404? there should always be a notification as response? Who will determine if the notification should be send or not?

src/main/java/it/gov/pagopa/payhub/activities/connector/ionotification/mapper/NotificationRequestMapper.java
```diff
@@ -1,30 +1,60 @@
...
 
 @Service
 @Lazy
 public class NotificationRequestMapper {
 
-    public List<NotificationRequestDTO> mapDebtPositionDTO2NotificationRequestDTO(DebtPositionDTO debtPosition, String serviceId, String subject, String markdown) {
-        return debtPosition.getPaymentOptions().stream()
+    public List<NotificationRequestDTO> map(List<PaymentOptionDTO> paymentOptions, Long orgId, Long debtPositionTypeOrgId, String apiKey, IONotificationDTO ioNotificationDTO, NotificationRequestDTO.OperationTypeEnum operationType) {
+        // If only one PaymentOption exists, map with nav field the first installment
+        if (paymentOptions.size() == 1) {
```
antonioT90, line 19: If single PO it cannot have more debtors? Otherwise let's always perform the stream and distinct operation

src/main/java/it/gov/pagopa/payhub/activities/exception/IONotificationInvocationException.java
```diff
@@ -0,0 +1,8 @@
+package it.gov.pagopa.payhub.activities.exception;
+
+public class IONotificationInvocationException extends RuntimeException{
```
antonioT90, line 3: Neither it's a NotRetryableException, neither you are catching it to handle it elsewhere...
Do we really need a customException?

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/connector/ionotification/client/IoNotificationClient.java
LarissaASLeite: No, here we could have RestClientException or exceptions that come from IO like IOWrongPayloadException.

The response could be null, but i was thinking about solve the null in the other PR with the new service.

I expected to be a separate activity called from workflow-hub, in this activity you will see the enum [IO, SEND] in the client of organization to get the apiKey

src/main/java/it/gov/pagopa/payhub/activities/exception/IONotificationInvocationException.java
LarissaASLeite: We don't necessarily need a custom exception. I was handling this in another PR, but I had some doubts here. I do think the exception should be retryable in this case, but I'm not sure which specific exception would be the best fit

src/main/java/it/gov/pagopa/payhub/activities/connector/ionotification/mapper/NotificationRequestMapper.java
robertopasetti: Potentially we can have multidebtor on the same PO, but at this time nodo pagoPA doesn't support this scenario. Given that in case we have a single PO we have to append notice info to the IO Message coming from the first installment, where the first installment is the installment with the minor due date (with UNPAID status)

src/main/java/it/gov/pagopa/payhub/activities/connector/ionotification/client/IoNotificationClient.java
antonioT90: If we expect that the service could return 404 to be transcoded into `null`, instead to be thrown like a new exception, let's handle it with a TODO with the JIRA ticket to remind to handle that `null`

antonioT90: I don't get the point on the enum with the enum, why should it have an impact here?

src/main/java/it/gov/pagopa/payhub/activities/connector/ionotification/mapper/NotificationRequestMapper.java
antonioT90: @robertopasetti this Activity is not related just to ACA, but it could be used in all pagoPa payments services.
Thus we should not think about "the nodo will not handle this shape of DP".

We could be more flexible and accept here the possibility to have more debtors on each TO_SYNC PO (we are building the notification considering just TO_SYNC Installments): this way we will handle all DP shape in the same manner instead of having a particular handle the PO is just 1 (or maybe we should always accept the debtor of each PO related to the earliest dueDate?)

src/main/java/it/gov/pagopa/payhub/activities/exception/IONotificationInvocationException.java
antonioT90: If the stacktrace of the possible Exception thrown and cacth here is printed somewhere, I would not introduce a custom Exception if we don't think to have a special handling for it.

All exceptions which go beyond the Activity is logged on Temporal and can be shown there: thus the information that the origin is the IONotification is already present: thus the origin of the exception is already visible.

It could be useful to explicit the point with a custom exception, specially if we expect that the executed code could easily gave an error (like when performing an integration).
Anyway, all REST integration are intercepted by `https://github.com/pagopa/p4pa-payhub-activities/blob/main/src/main/java/it/gov/pagopa/payhub/activities/config/RestTemplateConfig.java#L47` if the `printBodyError` is true (like in this case): thus we are already logging the particular API called as infrastructure service.

Thus I would suggest to remove this custom exception in order to "keep it simple" as stated in the development handbook

src/main/java/it/gov/pagopa/payhub/activities/connector/ionotification/client/IoNotificationClient.java
LarissaASLeite: It doesn't, i got wrong your first question. The field operationType will determine if the notification should be send or not, if (CREATE) send notification, this issue will be solved in the next [PR](https://pagopa.atlassian.net/browse/P4ADEV-2089)

src/main/java/it/gov/pagopa/payhub/activities/connector/ionotification/mapper/NotificationRequestMapper.java
antonioT90: @robertopasetti actually handling multi-debtor has not more effort: instead revert it yes because Larissa has already implemented it.
The only point is that the multi-debtor was not handled when the PO was just 1.

So I was asking why there was that misalignment based on the number of PO

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/connector/ionotification/mapper/NotificationRequestMapper.java
antonioT90: @LarissaASLeite let's handle the multi-debtor for the 1 only PO useCase as already done when there are multiple PO

antonioT90: @LarissaASLeite the logic should be:
* if just one PO: let's send as many notification as many Installments having the earliest dueDate (if defined) if neither Installment has a due Date, let's take the first
* if there are more than 1 PO, for each distinct Debtor, we should send a message without any reference to the IUV

is it right @robertopasetti?

LarissaASLeite: "if there are more than 1 PO, for each distinct Debtor, we should send a message without any reference to the IUV"

Without any reference to the NAV is more accurate

antonioT90 (approved):

PR #162, opened 2025-03-12
feat: P4ADEV-2397-add-service-to-calculate-operation-type

LarissaASLeite:

- Pre-Deploy Test
    - [ ] Unit
    - [ ] Integration (Narrow)
- Post-Deploy Test
    - [ ] Isolated Microservice
    - [ ] Broader Integration
    - [ ] Acceptance
    - [ ] Performance & Load

#### Types of changes

#### Checklist:


src/main/java/it/gov/pagopa/payhub/activities/connector/ionotification/mapper/NotificationRequestMapper.java
```diff
@@ -50,28 +52,41 @@ public List<NotificationRequestDTO> map(List<PaymentOptionDTO> paymentOptions, L
...
         }
         notificationRequestDTO.setAmount(installmentDTO.getAmountCents());
-        notificationRequestDTO.setOperationType(operationType);
+        notificationRequestDTO.setOperationType(NotificationRequestDTO.OperationTypeEnum.CREATE_DP);
 
         return notificationRequestDTO;
     }
+
+    private String replacePlaceholders(String markdown, InstallmentDTO installment) {
+        String placeholderDefault = "Not Specified";
```
antonioT90, line 78: @robertopasetti in case of value not available, we should set an empty string or a placeholder? in what language?

src/main/java/it/gov/pagopa/payhub/activities/connector/ionotification/mapper/NotificationRequestMapper.java
```diff
@@ -50,28 +52,41 @@ public List<NotificationRequestDTO> map(List<PaymentOptionDTO> paymentOptions, L
...
         notificationRequestDTO.setDebtPositionTypeOrgId(debtPositionTypeOrgId);
         if (ioNotificationDTO.getIoTemplateSubject() != null && ioNotificationDTO.getIoTemplateMessage() != null && ioNotificationDTO.getServiceId() != null) {
             notificationRequestDTO.setSubject(ioNotificationDTO.getIoTemplateSubject());
-            notificationRequestDTO.setMarkdown(ioNotificationDTO.getIoTemplateMessage());
+            notificationRequestDTO.setMarkdown(replacePlaceholders(ioNotificationDTO.getIoTemplateMessage(), installmentDTO));
             notificationRequestDTO.setServiceId(ioNotificationDTO.getServiceId());
         }
         notificationRequestDTO.setAmount(installmentDTO.getAmountCents());
-        notificationRequestDTO.setOperationType(operationType);
+        notificationRequestDTO.setOperationType(NotificationRequestDTO.OperationTypeEnum.CREATE_DP);
```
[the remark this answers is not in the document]
LarissaASLeite: The CREATE_DP value is used because, before calling the mapper, we determine the operationType using the `InstallmentOperationTypeResolver`. At this stage, we only send notifications for CREATE_DP (at least for now). This means that when we reach the mapper, CREATE_DP is already implicit in the process.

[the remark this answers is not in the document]
antonioT90: let's keep the value as input, ASAP we will introduce the need for the other operation in case of parameter "Execution Config" on the WF (as stated on the Confluence page which we discussed this morning)

src/main/java/it/gov/pagopa/payhub/activities/connector/ionotification/mapper/NotificationRequestMapper.java
```diff
@@ -7,17 +7,20 @@
...
 
+import static it.gov.pagopa.payhub.activities.util.Utilities.centsAmountToEuroString;
 import static java.util.stream.Collectors.groupingBy;
 
 @Service
 @Lazy
 public class NotificationRequestMapper {
 
-    public List<NotificationRequestDTO> map(List<PaymentOptionDTO> paymentOptions, Long orgId, Long debtPositionTypeOrgId, IONotificationDTO ioNotificationDTO, NotificationRequestDTO.OperationTypeEnum operationType) {
+    public List<NotificationRequestDTO> map(List<PaymentOptionDTO> paymentOptions, Long orgId, Long debtPositionTypeOrgId, String dpDescription, IONotificationDTO ioNotificationDTO) {
```
antonioT90, line 23: let's ask the entire DebtPosition instead of just the dpDescription!
This way we could add simply any new placeholder if needed

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/connector/ionotification/mapper/NotificationRequestMapper.java
```diff
@@ -50,28 +56,49 @@ public List<NotificationRequestDTO> map(List<PaymentOptionDTO> paymentOptions, L
...
+
+        placeholders.put("%posizioneDebitoria_descrizione%", Objects.toString(dpDescription, ""));
+        placeholders.put("%debitore_nomeCompleto%", Objects.toString(installment.getDebtor().getFullName(), ""));
+        placeholders.put("%debitore_codiceFiscale%", Objects.toString(installment.getDebtor().getFiscalCode(), ""));
+        placeholders.put("%importoTotale%", Objects.toString(centsAmountToEuroString(installment.getAmountCents()), ""));
+        placeholders.put("%IUV%", Objects.toString(installment.getIuv(), ""));
+        placeholders.put("%NAV%", Objects.toString(installment.getNav(), ""));
+        placeholders.put("%causale%", Objects.toString(installment.getRemittanceInformation(), ""));
+
+        String formattedDueDate = installment.getDueDate() != null ? installment.getDueDate().format(DateTimeFormatter.ofPattern("dd/MM/yyyy")) : "";
```
antonioT90, line 92: let's move the placeHolder configuration on a separate class in order to immediate locate it in case of new placeHolder configuration.

Set the DateTimeFormtter as a class constant, it will be probably re-used (keep it public)

antonioT90 (changes requested):

antonioT90 (dismissed):

antonioT90 (approved):

PR #168, opened 2025-03-14
feat: P4ADEV-2395 added dto and mapper

Giuseppe-LaManna:
#### List of Changes

-  added dto and mapper

- Pre-Deploy Test
    - [ ] Unit
    - [ ] Integration (Narrow)
- Post-Deploy Test
    - [ ] Isolated Microservice
    - [ ] Broader Integration
    - [ ] Acceptance
    - [ ] Performance & Load

#### Types of changes

#### Checklist:


src/main/java/it/gov/pagopa/payhub/activities/enums/EntityIdentifierType.java
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    public static EntityIdentifierType fromValue(String value) {
+        for (EntityIdentifierType b : EntityIdentifierType.values()) {
+            if (b.value.equals(value)) {
+                return b;
+            }
+        }
+        throw new IllegalArgumentException("Unexpected value '" + value + "'");
+    }}
```
antonioT90, line 24: if name and value are always equals, all of these is not needed.
Instead, let's add a javadoc to explain values (it's not obvious)
```suggestion
    /** Phisical person */
    F,
    /** Giuridical person */
    G;
}
```

src/main/java/it/gov/pagopa/payhub/activities/enums/UniqueIdentifierType.java
```diff
@@ -0,0 +1,7 @@
+package it.gov.pagopa.payhub.activities.enums;
+
+public enum UniqueIdentifierType {
```
antonioT90, line 3: can we add javadoc on these values?

src/main/java/it/gov/pagopa/payhub/activities/dto/export/debtposition/InstallmentExportFlowFileDTO.java
```diff
@@ -0,0 +1,169 @@
...
+import java.math.BigDecimal;
+import java.time.OffsetDateTime;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+public class InstallmentExportFlowFileDTO {
+
+    @CsvBindByName(column = "IUF")
```
antonioT90, line 20: Use `profiles` property to define the version of the track which support the field.
When configuring the CsvReader you will have to set the profile related to the track version.
Let's introduce a constant for each version in order to centralize the profile value (which will result to be aligned with track version)
EXPORT_PAID_VERSION_V1
EXPORT_PAID_VERSION_V1_1

Let's introduce a constant on String[] which will group the common profiles grouped:
EXPORT_PAID_INEHERITED_V1 = [V1, V1_1, V1_2]
EXPORT_PAID_INEHERITED_V1_1 = [V1_1, V1_2]
...

src/main/java/it/gov/pagopa/payhub/activities/dto/export/debtposition/InstallmentExportFlowFileDTO.java
```diff
@@ -0,0 +1,169 @@
...
+import lombok.NoArgsConstructor;
+
+import java.math.BigDecimal;
+import java.time.OffsetDateTime;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+public class InstallmentExportFlowFileDTO {
```
antonioT90, line 18:
```suggestion
public class PaidInstallmentExportFlowFileDTO {
```

src/main/java/it/gov/pagopa/payhub/activities/mapper/exportflow/debtposition/InstallmentExportFlowFileDTOMapper.java
```diff
@@ -0,0 +1,126 @@
...
+import it.gov.pagopa.pu.debtposition.dto.generated.Person;
+import org.springframework.context.annotation.Lazy;
+import org.springframework.stereotype.Service;
+
+@Lazy
+@Service
+public class InstallmentExportFlowFileDTOMapper {
+
+    public InstallmentExportFlowFileDTO map(Float versionTrack, InstallmentPaidViewDTO installmentPaidViewDTO){
+        switch (versionTrack.toString()){
```
antonioT90, line 17: due to the usage of the CSV profile utility, we could avoid this:
we could always fill all fields, it will be the CSV profile to handle the writing on the file or not.

just for the data which require other integrations (see RT field) which could introduce an if to explicitly set if the version will require it

src/main/java/it/gov/pagopa/payhub/activities/dto/export/debtposition/InstallmentExportFlowFileDTO.java
antonioT90: ps: it's more than a suggestion: the presence of the field will create the column: you was creating CSV for older version with the new columns, simply they were null

src/main/java/it/gov/pagopa/payhub/activities/mapper/exportflow/debtposition/InstallmentExportFlowFileDTOMapper.java
```diff
@@ -0,0 +1,126 @@
...
+        InstallmentExportFlowFileDTO installmentExportFlowFileDTO = mapToInstallmentExportFlowFileDTO1(installmentPaidViewDTO);
+
+        installmentExportFlowFileDTO.setPaymentReason(installmentPaidViewDTO.getRemittanceInformation());
+        installmentExportFlowFileDTO.setCollectionSpecificData("9/".concat(installmentPaidViewDTO.getCategory()));
+        installmentExportFlowFileDTO.setDueType(installmentPaidViewDTO.getCode());
+        installmentExportFlowFileDTO.setRt(null); //TODO field rt depends on task https://pagopa.atlassian.net/browse/P4ADEV-2306
+        installmentExportFlowFileDTO.setSinglePaymentDataIndex(installmentPaidViewDTO.getTransferIndex());
+        installmentExportFlowFileDTO.setPspAppliedFees(installmentPaidViewDTO.getFeeCents() != null ? Utilities.longCentsToBigDecimalEuro(installmentPaidViewDTO.getFeeCents()) : null);
+
+        if (installmentPaidViewDTO.getCode().equals("MARCA_BOLLO")){
```
antonioT90, line 100: is it there a constant for this?

antonioT90 (changes requested):

src/test/java/it/gov/pagopa/payhub/activities/mapper/exportflow/debtposition/PaidInstallmentExportFlowFileDTOMapperTest.java
```diff
@@ -0,0 +1,202 @@
...
+    void  givenValidInstallmentPaidView_whenMap_thenReturnInstallmentPaidViewDTO() {
+        //given
+        InstallmentPaidViewDTO installmentPaidViewDTO = podamFactory.manufacturePojo(InstallmentPaidViewDTO.class);
+        installmentPaidViewDTO.setCode("MARCA_BOLLO");
+        //when
+        PaidInstallmentExportFlowFileDTO result = installmentExportFlowFileDTOMapper.map(installmentPaidViewDTO);
+        //then
+        assertNotNull(result);
+        assertAllField(installmentPaidViewDTO, result);
+        TestUtils.checkNotNullFields(result,
```
antonioT90, line 37: all of these fields could be null?

src/main/java/it/gov/pagopa/payhub/activities/dto/export/debtposition/PaidInstallmentExportFlowFileDTO.java
```diff
@@ -0,0 +1,173 @@
...
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+public class PaidInstallmentExportFlowFileDTO {
+    private static final String EXPORT_PAID_VERSION_V1= "v1.0";
+    private static final String EXPORT_PAID_VERSION_V1_1= "v1.1";
+    private static final String EXPORT_PAID_VERSION_V1_2= "v1.2";
+    private static final String EXPORT_PAID_VERSION_V1_3= "v1.3";
+
+    @CsvBindByName(column = "IUF", profiles = EXPORT_PAID_VERSION_V1 )
```
antonioT90, line 24: you have to explicit report all versions which should write this column!
also for the other fields

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/dto/export/debtposition/PaidInstallmentExportFlowFileDTO.java
```diff
@@ -0,0 +1,187 @@
...
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+public class PaidInstallmentExportFlowFileDTO {
+    private static final String EXPORT_PAID_VERSION_V1= "v1.0";
+    private static final String EXPORT_PAID_VERSION_V1_1= "v1.1";
+    private static final String EXPORT_PAID_VERSION_V1_2= "v1.2";
+    private static final String EXPORT_PAID_VERSION_V1_3= "v1.3";
```
antonioT90, line 23: Let's create a "ExportConstants" class and move these there.
When we will introduce the exportTrackingVersion (or whetever will be its name) on the ExportFile entity, we will validate its value based on the requested ExportFileType, thus probably we will next introduce a Map<ExportFileType, Set<String>> to collect the available versions based on the file type requested

antonioT90: ps: we could already introduce the Map fill it with Paid requirements

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/dto/export/ExportConstants.java
```diff
@@ -0,0 +1,30 @@
...
+
+    private ExportConstants (){}
+
+    static {
+        Set<String> paidVersions = new HashSet<>();
+        paidVersions.add(ExportConstants.EXPORT_PAID_VERSION_V1);
+        paidVersions.add(ExportConstants.EXPORT_PAID_VERSION_V1_1);
+        paidVersions.add(ExportConstants.EXPORT_PAID_VERSION_V1_2);
+        paidVersions.add(ExportConstants.EXPORT_PAID_VERSION_V1_3);
+        availableVersions.put(ExportFile.FlowFileTypeEnum.PAID, paidVersions);
```
antonioT90, line 24: it's better an unmutable approach:

availableVersions = Map.of(
   ExportFile.FlowFileTypeEnum.PAID, Set.of(...)
);

antonioT90 (changes requested):

antonioT90 (approved):

PR #170, opened 2025-03-14
fix: P4ADEV-2089-fix-activity-io-notification

LarissaASLeite:

- Pre-Deploy Test
    - [ ] Unit
    - [ ] Integration (Narrow)
- Post-Deploy Test
    - [ ] Isolated Microservice
    - [ ] Broader Integration
    - [ ] Acceptance
    - [ ] Performance & Load

#### Types of changes

#### Checklist:


src/main/java/it/gov/pagopa/payhub/activities/service/debtposition/ionotification/IONotificationService.java
```diff
@@ -0,0 +1,112 @@
...
+                debtPositionDTO, ioNotificationDTO, CREATE_DP
+        );
+
+        return notifications.stream()
+                .map(ioNotificationFacadeService::sendMessage)
+                .toList();
+    }
+
+    /**
+     * Filters payment options to keep only installments with the CREATE_DP operation type.
```
antonioT90, line 73: this is wrong... it should no be the Activity to decide if notify or not just the CREATE.
it should the DebtPositionTypeOrg configuration to decide.
you should call the DebtPositionTypeOrg with the calculated operationType: if it return a config, you have to send the message, otherwise not

src/main/java/it/gov/pagopa/payhub/activities/service/debtposition/ionotification/IONotificationService.java
```diff
@@ -0,0 +1,112 @@
...
+     */
+    private DebtPositionDTO calculateOperationType(DebtPositionDTO debtPositionDTO) {
+        debtPositionDTO.getPaymentOptions().forEach(po ->
+                po.setInstallments(po.getInstallments().stream()
+                        .filter(i -> NotificationRequestDTO.OperationTypeEnum.CREATE_DP
+                                .equals(installmentOperationTypeResolver.calculateOperationType(i)))
```
antonioT90, line 79: This way you are deciding if it's a creation just considering the single installment, if you are adding an installment to an already existing DP this should be considered a CREATE operation?

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/service/debtposition/ionotification/IONotificationService.java
```diff
@@ -0,0 +1,70 @@
...
+        this.debtPositionTypeOrgService = debtPositionTypeOrgService;
+        this.ioNotificationFacadeService = ioNotificationFacadeService;
+        this.notificationRequestMapper = notificationRequestMapper;
+        this.debtOperationOperationTypeResolver = debtOperationOperationTypeResolver;
+    }
+
+    public List<MessageResponseDTO> sendMessage(DebtPositionDTO debtPositionDTO, Map<String, IupdSyncStatusUpdateDTO> iupdSyncStatusUpdateDTOMap) {
+        List<MessageResponseDTO> response = new ArrayList<>();
+
+        NotificationRequestDTO.OperationTypeEnum operationType = debtOperationOperationTypeResolver.calculateDebtPositionOperationType(debtPositionDTO, iupdSyncStatusUpdateDTOMap);
```
antonioT90, line 41: Is this enum really necessary to io-notification request?
It could be better to have a dedicated enum defined on this project as return type of `debtOperationOperationTypeResolver.calculateDebtPositionOperationType`.

Moreover, io-notification should accept the final message, the operation is really needed?

src/main/java/it/gov/pagopa/payhub/activities/service/debtposition/ionotification/IONotificationService.java
```diff
@@ -0,0 +1,70 @@
...
+
+        if (operationType != null) {
+            IONotificationDTO ioNotificationDTO = debtPositionTypeOrgService
+                    .getIONotificationDetails(debtPositionDTO.getDebtPositionTypeOrgId(), operationType);
+
+            if (ioNotificationDTO != null && operationType.equals(CREATE_DP)) {
```
antonioT90, line 47: if the notificaiton should be send or not, should be defined just from the presence of the IONotificationDTO, not the type of operation
```suggestion
            if (ioNotificationDTO != null) {
```

src/main/java/it/gov/pagopa/payhub/activities/service/debtposition/ionotification/DebtOperationOperationTypeResolver.java
```diff
@@ -0,0 +1,64 @@
...
+public class DebtOperationOperationTypeResolver {
+
+    private final InstallmentOperationTypeResolver installmentOperationTypeResolver;
+
+    public DebtOperationOperationTypeResolver(InstallmentOperationTypeResolver installmentOperationTypeResolver) {
+        this.installmentOperationTypeResolver = installmentOperationTypeResolver;
+    }
+
+    public NotificationRequestDTO.OperationTypeEnum calculateDebtPositionOperationType(DebtPositionDTO debtPositionDTO, Map<String, IupdSyncStatusUpdateDTO> iupdSyncStatusUpdateDTOMap) {
+
```
antonioT90, line 25: if iupdSyncStatusUpdateDTOMap.isEmpty you should return null -> no op performed, thus not notification

src/main/java/it/gov/pagopa/payhub/activities/service/debtposition/ionotification/DebtOperationOperationTypeResolver.java
```diff
@@ -0,0 +1,64 @@
...
+            } else if (operations.stream().allMatch(op -> op.equals(NotificationRequestDTO.OperationTypeEnum.DELETE_DP))) {
+                return NotificationRequestDTO.OperationTypeEnum.DELETE_DP;
+            } else {
+                return NotificationRequestDTO.OperationTypeEnum.UPDATE_DP;
+            }
+        }
+    }
+
+    public boolean isUpdateOperationType(DebtPositionDTO debtPositionDTO) {
+        // If not even one is TO_SYNC, INVALID o CANCELLED then is UPDATE_PD
```
antonioT90, line 55:
```suggestion
        // If exists at least one active (not CANCELLED or INVALID) Installment not TO_SYNC
```

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/service/debtposition/ionotification/DebtOperationOperationTypeResolver.java
antonioT90: the check on iupdSyncStatusUpdateDTOMap is to be performed as first!
it should be executed before to check if the DP has active Installments

antonioT90 (changes requested):

antonioT90 (dismissed):

antonioT90 (approved):

PR #190, opened 2025-03-21
feat: P4ADEV-2466 - TransferClassificationActivity - add field debtPositionTypeOrgCode 

macacia:

- Pre-Deploy Test
    - [x] Unit
    - [ ] Integration (Narrow)
- Post-Deploy Test
    - [ ] Isolated Microservice
    - [ ] Broader Integration
    - [ ] Acceptance
    - [ ] Performance & Load

#### Types of changes

#### Checklist:

- [x] I have updated the documentation accordingly.


src/main/java/it/gov/pagopa/payhub/activities/connector/debtposition/DebtPositionTypeOrgServiceImpl.java
```diff
@@ -42,4 +42,15 @@ public IONotificationDTO getDefaultIONotificationDetails(Long debtPositionTypeOr
             return null;
         }
     }
+
+    @Override
+    public DebtPositionTypeOrg getDebtPositionTypeOrgByInstallmentId(Long installmentId) {
+        log.info("Fetching DebtPositionTypeOrg with installmentId: {}", installmentId);
+        try {
```
antonioT90, line 49: here I expect you have nothing to catch...
why should you transform a generic exception into null?

antonioT90 (changes requested):

antonioT90 (approved):

PR #302, opened 2025-05-30
feat: P4ADEV-2992-activity-import-pagati

RiccardoGiuliani:

- Pre-Deploy Test
    - [ ] Unit
    - [ ] Integration (Narrow)
- Post-Deploy Test
    - [ ] Isolated Microservice
    - [ ] Broader Integration
    - [ ] Acceptance
    - [ ] Performance & Load

#### Types of changes

#### Checklist:


src/main/java/it/gov/pagopa/payhub/activities/dto/ingestion/receipt/ReceiptIngestionFlowFileDTO.java
antonioT90, line 1: let's use english for property name, the italian name should be only on the openCsv annotations.
let's give as field name something related to the returned data (use flowSourceName, or sourceFlowName, instead of iuf)

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/receipt/ReceiptProcessingService.java
```diff
@@ -0,0 +1,94 @@
...
+     * @return An {@link ReceiptIngestionFlowFileResult} containing details about the processed rows, errors, and archived files.
+     */
+    public ReceiptIngestionFlowFileResult processReceipts(Iterator<ReceiptIngestionFlowFileDTO> iterator,
+                                                          List<CsvException> readerExceptions,
+                                                          IngestionFlowFile ingestionFlowFile,
+                                                          Path workingDirectory) {
+        List<ReceiptErrorDTO> errorList = new ArrayList<>();
+        ReceiptIngestionFlowFileResult result = new ReceiptIngestionFlowFileResult();
+        process(iterator, readerExceptions, result, ingestionFlowFile, errorList, workingDirectory);
+        result.setFileVersion(ingestionFlowFile.getFileVersion());
```
antonioT90, line 53: this is wrong, just the import of DP_INSTALLMENTS knows during import the version, here you should determine it based on the profile configured during the read.
Actually, let's set it to 1.0
```suggestion
        result.setFileVersion("1.0");
```

antonioT90 (changes requested):

PR #355, opened 2025-07-15
feat: P4ADEV-3224 service id on debt position type org

domenicogi:

- Pre-Deploy Test
    - [X] Unit
    - [ ] Integration (Narrow)
- Post-Deploy Test
    - [ ] Isolated Microservice
    - [ ] Broader Integration
    - [ ] Acceptance
    - [ ] Performance & Load

#### Types of changes

#### Checklist:

### DIFF
--- a/src/main/java/it/gov/pagopa/payhub/activities/mapper/ingestionflow/debtpositiontypeorg/DebtPositionTypeOrgMapper.java
+++ b/src/main/java/it/gov/pagopa/payhub/activities/mapper/ingestionflow/debtpositiontypeorg/DebtPositionTypeOrgMapper.java
@@ -42,18 +42,20 @@ public class DebtPositionTypeOrgMapper {
         .flagNotifyOutcomePush(dto.isFlagNotifyOutcomePush())
         .flagAmountActualization(dto.isFlagAmountActualization())
         .flagExternal(dto.isFlagExternal())
-        .notifyOutcomePushOrgSilServiceId(getOrgSilServiceId(organizationId, OrgSilServiceType.PAID_NOTIFICATION_OUTCOME))
-        .amountActualizationOrgSilServiceId(getOrgSilServiceId(organizationId, OrgSilServiceType.ACTUALIZATION))
+        .notifyOutcomePushOrgSilServiceId(getOrgSilServiceId(organizationId, OrgSilServiceType.PAID_NOTIFICATION_OUTCOME,dto.getNotifyOutcomePushOrgSilServiceCode()))
+        .amountActualizationOrgSilServiceId(getOrgSilServiceId(organizationId, OrgSilServiceType.ACTUALIZATION,dto.getAmountActualizationOrgSilServiceCode()))
+        .serviceId(dto.getServiceCode())
         .build();
   }
 
 
-  private Long getOrgSilServiceId(Long organizationId, OrgSilServiceType orgSilServiceType) {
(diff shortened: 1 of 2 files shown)


src/main/java/it/gov/pagopa/payhub/activities/mapper/ingestionflow/debtpositiontypeorg/DebtPositionTypeOrgMapper.java
```diff
@@ -33,8 +42,22 @@ public DebtPositionTypeOrgRequestBody map(DebtPositionTypeOrgIngestionFlowFileDT
         .flagNotifyOutcomePush(dto.isFlagNotifyOutcomePush())
         .flagAmountActualization(dto.isFlagAmountActualization())
         .flagExternal(dto.isFlagExternal())
+        .notifyOutcomePushOrgSilServiceId(getOrgSilServiceId(organizationId, OrgSilServiceType.PAID_NOTIFICATION_OUTCOME))
```
antonioT90, line 45: should the DTO define the applicationName of the OrgSilService?

src/test/java/it/gov/pagopa/payhub/activities/mapper/ingestionflow/debtpositiontypeorg/DebtPositionTypeOrgMapperTest.java
```diff
@@ -27,12 +37,23 @@ void givenDebtPositionTypeIngestionFlowFileWithPodamFactoryWhenMapThenOk() {
...
     var result = debtPositionTypeMapper.map(dto, debtPositionTypeId, organizationId);
 
     Assertions.assertNotNull(result);
+    Assertions.assertEquals(111L, result.getNotifyOutcomePushOrgSilServiceId());
+    Assertions.assertEquals(222L, result.getAmountActualizationOrgSilServiceId());
     checkNotNullFields(result, "creationDate", "updateDate", "updateOperatorExternalId",
-        "updateTraceId", "debtPositionTypeOrgId","serviceId","notifyOutcomePushOrgSilServiceId","amountActualizationOrgSilServiceId");
-    //todo : remove exclusion in checkNotNullFields for serviceId, notifyOutcomePushOrgSilServiceId e amountActualizationOrgSilServiceId with task https://pagopa.atlassian.net/browse/P4ADEV-3224
+        "updateTraceId", "debtPositionTypeOrgId", "serviceId");
+    // ora i campi notifyOutcomePushOrgSilServiceId e amountActualizationOrgSilServiceId sono valorizzati
```
antonioT90, line 56: please, remove this italian comment (not usefull to translate it into english)

os: why should serviceId null? this is the IO serviceId... why are we not accepting it?

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/mapper/ingestionflow/debtpositiontypeorg/DebtPositionTypeOrgMapper.java
marcomatteuccieng: Yes, to retrieve the related org_sil_service_id you need to use the applicationName field.
I noticed that it will no longer be possible to have two rows in the org_sil_service table with the same applicationName  but different URLs associated with different debt_position_type_org, whereas this was possible in MyPay4.

antonioT90 (approved):

PR #400, opened 2025-09-03
fix: P4ADEV-0001 fix iud classification

domenicogi:

- Pre-Deploy Test
    - [X] Unit
    - [ ] Integration (Narrow)
- Post-Deploy Test
    - [ ] Isolated Microservice
    - [ ] Broader Integration
    - [ ] Acceptance
    - [ ] Performance & Load

#### Types of changes

#### Checklist:


src/main/java/it/gov/pagopa/payhub/activities/connector/organization/TaxonomyServiceImpl.java
```diff
@@ -23,4 +27,11 @@ public Integer syncTaxonomies() {
 		log.debug("Synchronizing taxonomies");
 		return taxonomyClient.fetchTaxonomies(authnService.getAccessToken());
 	}
+
+	@Override
+	@Cacheable(key = "#organizationType + '-' + #macroAreaCode + '-' + #serviceTypeCode + '-' + #collectionReason",
+			unless = "#result.getEmbedded().getTaxonomies().size() == 0")
```
antonioT90, line 33: on activities the caching behavior is not enabled and configured! you cannot use this

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/debtpositiontype/DebtPositionTypeProcessingService.java
```diff
@@ -94,6 +98,31 @@ protected boolean consumeRow(long lineNumber,
         return false;
       }
 
+      String category = debtPositionTypeDTO.getTaxonomyCode();
```
antonioT90, line 101: why are you performing this check here? validation checks are supposed to be on the API, not on the CSV reading...

@Benedetta-fabbri have we implemented such control inside debt-positions?

antonioT90 (changes requested):

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/debtpositiontype/DebtPositionTypeProcessingService.java
Benedetta-fabbri: Yes, but this is a validation for transfer category inside ValidateDebtPositionServiceImpl. Maybe we can create an API in p4pa-organization that validate taxonomy and we can use it in p4pa-debt-positions and here

src/main/java/it/gov/pagopa/payhub/activities/connector/organization/TaxonomyServiceImpl.java
domenicogi: removed

src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/debtpositiontype/DebtPositionTypeProcessingService.java
antonioT90: @domenicogi I created a task to implement this validation on the API: https://pagopa.atlassian.net/browse/P4ADEV-3775

let's remove the check here.

In general, we should implement all the validation on the API and not on the CSV reading! This way the check will be centralized and the API will return the 400 with the error describe which will describe the contraint violated

antonioT90 (approved):

PR #401, opened 2025-09-08
fix: P4ADEV-0005 add check to iuf already associated to another treasury

domenicogi:

- Pre-Deploy Test
    - [X] Unit
    - [ ] Integration (Narrow)
- Post-Deploy Test
    - [ ] Isolated Microservice
    - [ ] Broader Integration
    - [ ] Acceptance
    - [ ] Performance & Load

#### Types of changes

#### Checklist:


src/main/java/it/gov/pagopa/payhub/activities/service/ingestionflow/treasury/csvcomplete/TreasuryCsvCompleteProcessingService.java
```diff
@@ -59,15 +61,40 @@ public TreasuryIufIngestionFlowFileResult processTreasuryCsvComplete(
...
                         row.getOrganizationIpaCode(), ipa);
                 log.error(errorMessage);
                 throw new OrganizationIpaCodeNotMatchException(errorMessage);
             }
+
+            Organization organization = organizationService.getOrganizationByIpaCode(ipa).orElseThrow(() -> {
+                String errorMessage = String.format("No organization found for IPA code %s", ipa);
+                log.error(errorMessage);
+                return new OrganizationIpaCodeNotMatchException(errorMessage);
+            });
```
antonioT90, line 78: this cannot happen! at row 66 we are checking ig the IPA code match the ingestionFlowFile organization... thus it should already be checked...
if you need organizationId, add it to result as done for ipa code instead of calling the REST service for each row

antonioT90 (changes requested):

antonioT90 (approved):

